PR #6, opened 2023-10-08
Added GitHub actions pipelines for building and releasing

gavlyukovskiy:


gradle.properties
```diff
@@ -0,0 +1 @@
+version=0.1.0-SNAPSHOT
```
gavlyukovskiy, line 1: Had to add it here, otherwise `project.version` is set to `unspecified` by default and I was reluctant to depend on that arbitrary string in the code

.github/workflows/build.yml
```diff
@@ -0,0 +1,76 @@
...
+    paths-ignore:
+      - README.adoc
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
gavlyukovskiy, line 23: This makes sure to fetch all tags as well, so that we can generate snapshot version correctly

Breus (approved):

PR #23, opened 2023-12-20
Add a regex benchmark for comparison

gavlyukovskiy:
Had a thought to check how bad regex is doing, I'm honestly surprised it's doing quite okay (only 120-200 times slower than json-masker) 😄 

```text
Benchmark                                                             (jsonSize)  (maskedKeyProbability)  (obfuscationLength)   Mode  Cnt        Score   Error  Units
JsonMaskerBenchmark.baselineCountBytes         1kb                    0.01                 none  thrpt       3196601,040          ops/s
JsonMaskerBenchmark.baselineCountBytes       128kb                    0.01                 none  thrpt         23401,364          ops/s
JsonMaskerBenchmark.baselineCountBytes         2mb                    0.01                 none  thrpt          1462,554          ops/s
JsonMaskerBenchmark.jacksonString              1kb                    0.01                 none  thrpt         91196,798          ops/s
JsonMaskerBenchmark.jacksonString            128kb                    0.01                 none  thrpt           491,485          ops/s
JsonMaskerBenchmark.jacksonString              2mb                    0.01                 none  thrpt            18,792          ops/s
JsonMaskerBenchmark.jsonMaskerBytes            1kb                    0.01                 none  thrpt        983466,058          ops/s
JsonMaskerBenchmark.jsonMaskerBytes          128kb                    0.01                 none  thrpt          5229,711          ops/s
JsonMaskerBenchmark.jsonMaskerBytes            2mb                    0.01                 none  thrpt           297,844          ops/s
JsonMaskerBenchmark.jsonMaskerString           1kb                    0.01                 none  thrpt        824710,487          ops/s
JsonMaskerBenchmark.jsonMaskerString         128kb                    0.01                 none  thrpt          3697,972          ops/s
JsonMaskerBenchmark.jsonMaskerString           2mb                    0.01                 none  thrpt           279,800          ops/s
JsonMaskerBenchmark.regexList                  1kb                    0.01                 none  thrpt         40465,063          ops/s
JsonMaskerBenchmark.regexList                128kb                    0.01                 none  thrpt            42,897          ops/s
JsonMaskerBenchmark.regexList                  2mb                    0.01                 none  thrpt             2,307          ops/s
```

### DIFF
--- a/src/jmh/benchmark-history/2023-12-20.md
+++ b/src/jmh/benchmark-history/2023-12-20.md
@@ -1,0 +1,88 @@
+- OS: Windows 11
+- CPU: AMD Ryzen 5 7600X
+- Java: 17.0.7
+
+```text
+Benchmark                                  (characters)  (jsonSize)  (maskedKeyProbability)  (obfuscationLength)   Mode  Cnt        Score   Error  Units
+BaselineBenchmark.countBytes                        N/A         1kb                     N/A                  N/A  thrpt       4074057.577          ops/s
+BaselineBenchmark.countBytes                        N/A       128kb                     N/A                  N/A  thrpt         33992.898          ops/s
+BaselineBenchmark.countBytes                        N/A         2mb                     N/A                  N/A  thrpt          2253.498          ops/s
+BaselineBenchmark.jacksonParseAndMask               N/A         1kb                     N/A                  N/A  thrpt        195683.419          ops/s
+BaselineBenchmark.jacksonParseAndMask               N/A       128kb                     N/A                  N/A  thrpt           722.508          ops/s
+BaselineBenchmark.jacksonParseAndMask               N/A         2mb                     N/A                  N/A  thrpt            27.133          ops/s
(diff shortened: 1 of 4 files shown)


gavlyukovskiy:
Need to redo benchmarks on my PC to make sure values are consistent

Breus:
@gavlyukovskiy 120-200 times slower with no support for array and object masking which are by far the most expensive parts of masking 😄

gavlyukovskiy:
@Breus I've also added benchmark parameter that uses different character sets and results are quite bad - every time we do any array resize (have quotes, use unicode or obfuscation length) then performance drops below regexes  :(

Breus:
@gavlyukovskiy we could add an "ultra-performant" configuration which doesn't resize at all for escaped quotes and unicode characters but replaces them by the same length mask in bytes. This could also be something we add in a later version of the library.

For the rest: The fact that obfuscation slows down the algorithm and is slower than a regex without that seems logical and not something to address: if you would do the same with a regex/jackson, it would quite likely be much slower than this library.

PR #27, opened 2023-12-23
Finished failure fuzzing testing and fixed a bug in KeyContainsMasker

Breus:


build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
     id 'maven-publish'
     id 'signing'
     id 'io.github.gradle-nexus.publish-plugin' version '1.3.0'
-    id 'com.adarshr.test-logger' version '4.0.0'
+//    id 'com.adarshr.test-logger' version '4.0.0'
```
gavlyukovskiy, line 7:
```suggestion
    id 'com.adarshr.test-logger' version '4.0.0'
```
Let's uncomment this, I did it to verify that weird output isn't caused by it, but forgot to turn it back

gavlyukovskiy (approved):

PR #28, opened 2023-12-28
Created Fuzzing against Jackson and fixed multiple small bugs found with ti

Breus:


src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -34,11 +38,14 @@ public final class KeyContainsMasker implements JsonMasker {
...
     public KeyContainsMasker(JsonMaskingConfig maskingConfig) {
-        this.allowMode = maskingConfig.getTargetKeyMode() == JsonMaskingConfig.TargetKeyMode.ALLOW;
         this.maskingConfig = maskingConfig;
         this.targetKeysTrie = new ByteTrie(!maskingConfig.caseSensitiveTargetKeys());
         for (String key : maskingConfig.getTargetKeys()) {
-            this.targetKeysTrie.insert(key);
+            if (maskingConfig.caseSensitiveTargetKeys()) {
+                this.targetKeysTrie.insert(key);
+            } else {
+                this.targetKeysTrie.insert(key.toLowerCase());
```
gavlyukovskiy, line 47: This shouldn't be necessary - trie already converts to lowercase/uppercase on prowler character basis, did it not work?

gavlyukovskiy (approved):
Left one comment, but otherwise LGTM :+1:

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: @gavlyukovskiy I think you are right, I just did this to replace:

```java
        if (!caseSensitiveTargetKeys) {
            targets = targets.stream().map(String::toLowerCase).collect(Collectors.toSet());
        }
        targetKeys = targets;
 ```
        
        Probably you forgot to remove that while introducing ByteTrie?

PR #35, opened 2024-01-03
Add Sonarqube codecoverage

gavlyukovskiy:


gavlyukovskiy:
Turns out it's not that easy. I need to pass a sonar token and switch from automated analysis to CI analysis in SonarCloud

PR #50, opened 2024-02-11
Unify @ParametersAreNonnullByDefault usage for all package

Breus:


gavlyukovskiy:
@Breus what do you think about doing it like spring? They have [custom annotations](https://github.com/spring-projects/spring-framework/tree/main/spring-core/src/main/java/org/springframework/lang) that cover both parameters and also return values ([usage example](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/util/package-info.java)).

gavlyukovskiy (approved):

PR #53, opened 2024-02-17
Add benchmark pipeline

gavlyukovskiy:
This PR adds a benchmark workflow that measures difference between latest commit of the PR and master.
The comment will be automatically updated when new commits are pushed to the PR to avoid spam.


Breus (approved):
This is super cool! I love it

PR #54, opened 2024-02-19
Added white space injected no-failed-execution fuzzing test

Breus:


src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -119,13 +119,18 @@ public Map<String, KeyMaskingConfig> getKeyConfigs() {
...
     @Override
     public String toString() {
-        return "JsonMaskingConfig{" +
-               "targetKeys=" + targetKeys +
-               ", targetKeyMode=" + targetKeyMode +
-               ", caseSensitiveTargetKeys=" + caseSensitiveTargetKeys +
-               ", defaultConfig=" + defaultConfig +
-               ", targetKeyConfigs=" + targetKeyConfigs +
-               '}';
+        return "JsonMaskingConfig{"
```
gavlyukovskiy, line 122: Multi-line string + `.formatted()`?

src/test/java/randomgen/json/RandomJsonWhiteSpaceInjector.java
```diff
@@ -0,0 +1,126 @@
...
+                && AsciiJsonUtil.isNumericCharacter(previousOrignalByte)) {
+            return false;
+        }
+        // Not allowed in boolean values (true/false)
+        if (isLowercaseA(originalJsonByte)
+                || isLowercaseL(originalJsonByte)
+                || isLowercaseS(originalJsonByte)
+                || isLowercaseE(originalJsonByte)
+                || isLowercaseR(originalJsonByte)
+                || isLowercaseU(originalJsonByte)) {
```
gavlyukovskiy, line 117: This is quite cryptic 😅

src/test/java/randomgen/json/RandomJsonWhiteSpaceInjector.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param jsonBytes the JSON bytes for which it will be checked if the whitespace can be
+     *     injected
+     * @param index the index for which it will be checked, must be > 1
+     * @return whether injecting a whitespace before the index in the provided JSON bytes will
+     *     result in valid JSON
+     */
+    private static boolean canInjectWhiteSpaceBeforeByteAtIndex(byte[] jsonBytes, int index) {
+        if (index < 1) {
+            throw new IllegalArgumentException(
```
gavlyukovskiy, line 97: Can't you inject a whitespace before object start?

src/test/java/randomgen/json/RandomJsonWhiteSpaceInjector.java
Breus: Hahaha I knew it when I wrote it 😅 
The comment on top of it that you cannot add whitespace inside boolean values makes it clear I think :-P

src/test/java/randomgen/json/RandomJsonWhiteSpaceInjector.java
Breus: You can, but honestly I don't really care about it. I just wanted to make sure there will always be a previous byte (I was only interested in whitespaces inside the JSON)

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
Breus: Let me give that a try

Breus: Updated

src/test/java/randomgen/json/RandomJsonWhiteSpaceInjector.java
Breus: Actually, I updated it so it is possible now.

src/test/java/randomgen/json/RandomJsonWhiteSpaceInjector.java
gavlyukovskiy: Doesn't it miss `t`? Also, it would be slightly better if you'd arrange the order of checks to be something like t, r, u, e, f, a, l, s

gavlyukovskiy: Or is it just starting from second byte?

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -119,13 +119,15 @@ public Map<String, KeyMaskingConfig> getKeyConfigs() {
...
-               ", targetKeyConfigs=" + targetKeyConfigs +
-               '}';
+        return """
+                JsonMaskingConfig{\n
+                targetKeys=%s,\n
+                targetKeyMode=%s,\n
+                caseSensitiveTargetKeys=%s,\n
+                defaultConfig=%s,\n
+                targetKeyConfigs=%s\n
+                }"""
```
gavlyukovskiy, line 129: I don't think you need `\n` now, does it not create two line breaks? Also I think `"""` needs to be on the new line so that indent is calculated correctly

gavlyukovskiy (approved):
Left one last comment (unimportant really), but otherwise looks good to me 👍

src/test/java/randomgen/json/RandomJsonWhiteSpaceInjector.java
Breus: It is starting from second byte and already in order ;-) (first false, then true), but e is already done in false

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
Breus: Lemme print it to check and I'll fix.

Breus: Updated this as well

PR #55, opened 2024-02-19
Adjust pipelines to be able to release pre-release versions

gavlyukovskiy:
Hope this pile of bash works :) Tested it locally by trying different versions.

P.S. I don't really know whether sonatype accepts our pre-release version in their central repo, but we'll try 😅

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,22 +42,30 @@ jobs:
         name: Generate snapshot version
         shell: bash
         run: |
+          # expects git tag to be in format 'v0.1.0' or 'v0.1.0-rc1'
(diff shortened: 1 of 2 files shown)


Breus (approved):
Bash is a delight to review as always

Breus:
> Hope this pile of bash works :) Tested it locally by trying different versions.
> 
> P.S. I don't really know whether sonatype accepts our pre-release version in their central repo, but we'll try 😅

@gavlyukovskiy I think they do, see for example: https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core/2.16.0-rc1

PR #56, opened 2024-02-21
Update readme to use latest API

gavlyukovskiy:


README.md
```diff
@@ -7,131 +7,330 @@
...
+var jsonMasker = JsonMasker.getMasker(
+        JsonMaskingConfig.builder()
+                .maskJsonPaths(Set.of("$.id", "$.nested.name"))
+                .build()
+);
+```
+
+Using `JsonMaskingConfig` allows customizing the masking behaviour of types, keys or json path or mix keys and json paths.
+
+> [!NOTE]
```
donavdey, line 75: Would also make sense to specify that jsonpath masking config takes precedent before simple key masking config (in a situation when jsonpath key and simple key point to the same key/value in the json)

donavdey (approved):

Breus (approved):

PR #57, opened 2024-02-21
Naive optimization of skipping any json path tracking when no json path keys are defined

gavlyukovskiy:
Not sure if we should go with this or get something more elaborate

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
@@ -74,7 +74,9 @@ public final class KeyContainsMasker implements JsonMasker {
         while (maskingState.currentIndex() < maskingState.messageLength() - MIN_OFFSET_JSON_KEY_QUOTE) {
             // Find JSON strings by looking for unescaped double quotes
             while (!currentByteIsUnescapedDoubleQuote(maskingState)) {
-                trackCurrentJsonPath(maskingState);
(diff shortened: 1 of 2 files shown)


Breus:
Oh lol 😂  Yea, I think we are aligned (saw this PR after the previous one)

@donavdey

gavlyukovskiy:
@Breus I think we might wait until state machine proposal is implemented. It would be much easier to do it there than having a (one more) layer of if statements. 
But this shows that we are getting ~30% slower on tracking the JsonPath on top of already losing about ~30% for supporting new API options.

donavdey:
We can still track current jsonpath, but skip querying the trie for the json path key as this is the most expensive operation. That is, on `trackJsonPath` option disabled always return an empty iterator over segments and avoid making `searchForJsonPathKey` call in general. What do you think? @gavlyukovskiy @Breus

gavlyukovskiy:
@donavdey it's actually tracking that takes the most time, not the search. I've been profiling it a couple of times and screenshot in https://github.com/Breus/json-masker/pull/58 also captured it. As you see there, it's roughly 30% for tracking and maybe 10% for trie lookup

donavdey:
@gavlyukovskiy I see now, this makes sense. I wonder if the track method can be optimized before focusing on the lookup optimization

gavlyukovskiy:
@breus @donavdey I'll close this one as it won't be compatible with #62. It was useful to know that we're loosing ~30% perf with JsonPath logic, so let's keep this in mind for future.

PR #58, opened 2024-02-22
Improve async-profiler integration

gavlyukovskiy:
This makes it easier to run benchmarks with async-profiler:
```
./gradlew jmh -PjmhShort -PjmhAsyncProfilerLibPath=/workspace/async-profiler/lib/libasyncProfiler.dylib
```
and then open the flamegraph in IDEA:
image

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -121,10 +121,9 @@ signing {
 }
 
 jmh {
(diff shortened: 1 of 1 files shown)


Breus:
Doesn't your screenshot show why we should indeed have JSONPath disabled in case no JSONPath is given as target? Since more than 25% of the CPU samples are in trackCurrentJsonPath 

@donavdey @gavlyukovskiy

PR #62, opened 2024-02-23
Support arrays in JSONPath keys

donavdey:
The PR adds support for json arrays in JSONPath keys and fixes the bug with root array and nested arrays in the root array not being masked.

The performance is expected to decrease as now the masker checks every array element if it needs to be masked. The check entails traversing the trie.

The optimization of keeping a trie node of the parent segment on the jsonpath dequeue is not yet implemented in this PR. I will implement it in the next PR.


src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -24,9 +25,9 @@ public final class KeyContainsMasker implements JsonMasker {
      */
     private static final int MIN_OFFSET_JSON_KEY_QUOTE = 3;
     /**
-     * Minimum JSON for which masking could be required is: {"":1}, so minimum length at least 6 bytes.
+     * Minimum JSON for which masking could be required is: [0], so minimum length at least 3 bytes.
      */
-    private static final int MIN_MASKABLE_JSON_LENGTH = 6;
+    private static final int MIN_MASKABLE_JSON_LENGTH = 3;
```
gavlyukovskiy, line 30: We might as well remove this "optimization" 😄

donavdey:
A small follow up: I made `isInArray` check to be the last operand in every "and" if statements in the track method and the benchmarks seem to report way smaller performance degradation.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Agreed. At this point it might as well go 😅 

However, I vaguely remember it had two purposes: the "optimization" but also protected against some ArrayIndexOutOfBound because it was reading ahead. 

I don't think that still applies, but lets make sure :-)

src/test/java/dev/blaauwendraad/masker/json/JsonPathArraysLookupTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.io.IOException;
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class JsonPathArraysLookupTest {
+    @ParameterizedTest
+    @MethodSource("jsonPathFile")
+    void test(JsonMaskerTestInstance testInstance) {
```
Breus, line 14: In general you don't have to name test instances something with test or even postfix them with `test` because it is already annotated with `@ParameterizedTest` and doesn't give much info while running. Something like `jsonPath` is slightly nicer

src/main/java/dev/blaauwendraad/masker/json/SegmentReference.java
```diff
@@ -0,0 +1,20 @@
...
+public class SegmentReference {
+    int start;
+    int offset;
+
+    SegmentReference(int start, int offset) {
+        this.start = start;
+        this.offset = offset;
+    }
+
+    public boolean isArraySegment() {
```
Breus, line 17: Maybe add some JavaDoc. Right now this is super clear to me, but I am afraid 1 year from now I might wonder what an `isArraySegment` means again

src/main/java/dev/blaauwendraad/masker/json/SegmentReference.java
```diff
@@ -0,0 +1,20 @@
+package dev.blaauwendraad.masker.json;
+
+/**
+ * A mutable reference to a sequence of bytes in <code>message</code>. It is used to represent json path segments.
+ * A "key" segment type reference is represented as a (start, offset) pair.
+ * For an "array index" segment type reference, a (start, offset) pair is interpreted as (index, -1).
+ */
+public class SegmentReference {
```
Breus, line 8: Can't this be called `JsonPathSegmentReference` or just `JsonPathSegment`? 

1 year from now I will not remember that SegmentReference is about JsonPath segments

src/main/java/dev/blaauwendraad/masker/json/SegmentReference.java
```diff
@@ -0,0 +1,20 @@
...
+    int start;
+    int offset;
```
Breus, line 10: Is it for performance to not make there private and use getters? I think that will be optimized by Hotspot anyway

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -104,23 +104,31 @@ public void backtrackCurrentJsonPath() {
...
-    }
-
     /**
      * Increments an array index in the last segment of the current jsonpath.
      * Throws {@link java.lang.IllegalStateException} if the last segment is not an array index.
      */
     public void incrementCurrentJsonPathArrayIndex() {
         SegmentReference lastSegment = currentJsonPath.peek();
-        assert isInArray() && lastSegment != null;
+        assert isInArraySegment() && lastSegment != null;
```
Breus, line 113: I think `assert`'s in production code is usually considered bad practice. 

Wouldn't it be better to just do `if (condition)`  with an IllegalStateException with a clear message what is going wrong? 

This might impact performance actually (we saw something like that in the past) so you might want to check benchmarks before/after

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -104,23 +104,31 @@ public void backtrackCurrentJsonPath() {
...
-     */
-    public boolean isInArray() {
-        return !currentJsonPath.isEmpty() && currentJsonPath.peek().offset == -1;
-    }
-
     /**
      * Increments an array index in the last segment of the current jsonpath.
      * Throws {@link java.lang.IllegalStateException} if the last segment is not an array index.
      */
     public void incrementCurrentJsonPathArrayIndex() {
```
Breus, line 111: Does this really need to be public?

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
Breus: Also: consider if we need this check in the first place: is it some internal code that calls this and can we make sure this never happens? If so, we might consider removing the check altogether

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -104,23 +104,31 @@ public void backtrackCurrentJsonPath() {
...
         SegmentReference lastSegment = currentJsonPath.peek();
-        assert isInArray() && lastSegment != null;
+        assert isInArraySegment() && lastSegment != null;
         lastSegment.start++;
     }
 
+    /**
+     * Checks if the current segment of the current jsonpath is an array index.
+     */
+    public boolean isInArraySegment() {
```
Breus, line 120: JavaDoc says "an array index", method says "in array segment"

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -173,6 +173,29 @@ public KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, int keyOffset, int
         }
     }
 
+    /**
+     * An overload of {@code dev.blaauwendraad.masker.json.KeyMatcher#getMaskConfigIfMatched(byte[], int, int, java.util.Iterator)} that does only jsonpath lookup.
+     */
```
Breus, line 178: > that does only jsonpath lookup.

=> "specifically for JsonPath look-ups?

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -173,6 +173,29 @@ public KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, int keyOffset, int
...
+    @CheckForNull
+    public KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, Iterator<SegmentReference> jsonPath) {
+        if (maskingConfig.isInMaskMode()) {
+            TrieNode node = searchForJsonPathKeyNode(bytes, jsonPath);
+            if (node != null && !node.negativeMatch) {
+                return node.keyMaskingConfig;
+            }
+            return null;
+        } else {
+            TrieNode node = searchForJsonPathKeyNode(bytes, jsonPath);
```
Breus, line 188: Maybe add one line of doc

// in allow mode

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -196,7 +219,7 @@ private TrieNode searchNode(byte[] bytes, int offset, int length) {
         return node;
     }
 
-    private TrieNode searchForJsonPathKeyNode(byte[] bytes, Iterator<MaskingState.SegmentReference> jsonPath) {
+    private TrieNode searchForJsonPathKeyNode(byte[] bytes, Iterator<SegmentReference> jsonPath) {
```
Breus, line 222: `@CheckForNull` on method return

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -207,9 +230,16 @@ private TrieNode searchForJsonPathKeyNode(byte[] bytes, Iterator<MaskingState.Se
             if (node == null) {
                 return null;
             }
-            MaskingState.SegmentReference segmentReference = jsonPath.next();
+            SegmentReference segmentReference = jsonPath.next();
             int keyStartIndex = segmentReference.start;
             int keyLength = segmentReference.offset;
+            if (segmentReference.isArraySegment()) {
+                node = node.children[Character.forDigit(keyStartIndex, 10) + BYTE_OFFSET];
```
Breus, line 237: `10` is a magic number. Please make it a named constant

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -69,6 +70,11 @@ public byte[] mask(byte[] input) {
         MaskingState maskingState = new MaskingState(input, 1);
         if (AsciiCharacter.isSquareBracketOpen(maskingState.byteAtCurrentIndexMinusOne())) {
             maskingState.expandCurrentJsonPath();
+            KeyMaskingConfig keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), maskingState.getCurrentJsonPath());
+            if (keyMaskingConfig != null) {
+                maskArrayElementInPlace(maskingState, keyMaskingConfig);
+                trackCurrentJsonPath(maskingState);
```
Breus, line 76: Since this is the main algorithm, it might be good to explain with comments left and right what is going on (like the rest of this file does whenever it is not obvious). This makes it easier for all of us to read the code and maintain it in the future

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -164,38 +161,103 @@ public byte[] mask(byte[] input) {
...
 
         ValueMaskingUtil.flushReplacementOperations(maskingState);
 
         return maskingState.getMessage();
     }
 
+    /**
+     * Masks the array element in place.
+     *
+     * @param maskingState the masking state an index prior to the element start index
```
Breus, line 174: "an index prior to the element start index" => "with the index on the element prior to the element start index"??

element start index is not super clear to me either, actually.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -164,38 +161,103 @@ public byte[] mask(byte[] input) {
...
+     */
+    private void maskArrayElementInPlace(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
+        maskingState.incrementCurrentIndex();
```
Breus, line 177: Wouldn't it make more sense to do this outside of calling this method and saying what you step over (one line of doc basically?)

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Actually it is only index prior the the start element index if there are no whitespaces (because those are skipped inside the method).

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -164,38 +161,103 @@ public byte[] mask(byte[] input) {
...
+    }
+
+    /**
+     * Masks the value in a key/value pair where the current index is on the value start
+     *
+     * @param maskingState     the current masking state in which the value will be masked
+     * @param keyMaskingConfig the masking configuration for the key
+     * @return true if the value is maskable, false otherwise
+     */
+    private boolean maskValue(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
```
Breus, line 193: Clever combination of masking and telling whether it was masked.

Might be good to change the JavaDoc description to include that clearly like: "Checks if the value on the current index in the MaskingState is maskable and masks it if so and return true. Otherwise, doesn't mask and returns false.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -164,38 +161,103 @@ public byte[] mask(byte[] input) {
...
+                && !keyMaskingConfig.isDisableBooleanMasking()) {
+            maskBooleanValueInPlace(maskingState, keyMaskingConfig);
+        } else {
+            return false;
+        }
+        return true;
+    }
+
     /**
      * Checks if the current byte changes the current json path. If so, updates current json path.
```
Breus, line 214: the current byte => 'the current byte in the maskingState` 

Also "changes" is not super clear from JavaDoc what this means

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -164,38 +161,103 @@ public byte[] mask(byte[] input) {
...
         // Check if this is the start of a json array
         if (currentByteIsUnescapedSquareBracketOpen(maskingState)) {
             maskingState.expandCurrentJsonPath();
+            KeyMaskingConfig keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), maskingState.getCurrentJsonPath());
+            if (keyMaskingConfig != null) {
+                maskArrayElementInPlace(maskingState, keyMaskingConfig);
+                trackCurrentJsonPath(maskingState);
+            }
+        }
+        if (currentByteIsUnescapedComma(maskingState)) {
```
Breus, line 247: Might be nice for clarity to add one line of JavaDoc explaining you are checking if there is a next element here.

Breus (commented):
The implementation looks actually quite nice (not too convoluted what I expected). 

It might be nice to sometimes add a bit more lines of doc in-between the algorithmic code to explain why some if/else is being done for others to not have to figure it out but so we can just read it like a story (like a lot of code in KeyContainsMasker has)

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: I will check if it can be removed

src/main/java/dev/blaauwendraad/masker/json/SegmentReference.java
```diff
@@ -0,0 +1,20 @@
...
+
+    SegmentReference(int start, int offset) {
+        this.start = start;
+        this.offset = offset;
+    }
+
+    public boolean isArraySegment() {
+        return this.offset == -1;
+    }
+}
```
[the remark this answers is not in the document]
donavdey: Makes sense, I will do it.

src/test/java/dev/blaauwendraad/masker/json/JsonPathArraysLookupTest.java
donavdey: I agree that "test" is implied from the context. I want to keep the method name a verb though, so I will come up with something in-between.

src/main/java/dev/blaauwendraad/masker/json/SegmentReference.java
donavdey: Makes sense, I will do it.

src/main/java/dev/blaauwendraad/masker/json/SegmentReference.java
donavdey: `JsonPathSegmentReference` is a more suitable name indeed. I will rename it.

src/main/java/dev/blaauwendraad/masker/json/SegmentReference.java
donavdey: Agree. I will make it a POJO bean.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: > is it some internal code that calls this and can we make sure this never happens

Exactly. I will remove it as If we are finished with adding more logic on top of the current masking loop. I am curious though about what would benchmarks say about "assertions" against "exceptions" performance.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Not really, I can make it package-private.
Ideally I would move all jsonpath-related logic into a separate "JsonPathTracker" class aggregated into MaskingState. But as we are going to start on token-based state machine that would not be necessary right now.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: I will fix it.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
donavdey: That is a better wording, I will use it.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
donavdey: I will add more docs there.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
donavdey: Ok.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: I will write more docs there.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: I basically took the description from a similar method. But I can rename it.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: It is good that all "mask" methods follow the same convention of starting one index prior to the value, so it would be a bit confusing if one method does not conform to it.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: Thanks! I will fix the javadoc.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: I will fix it.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: I will do it.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Makes sense

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -173,6 +174,30 @@ public KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, int keyOffset, int
         }
     }
 
+    /**
+     * An overload of {@code dev.blaauwendraad.masker.json.KeyMatcher#getMaskConfigIfMatched(byte[], int, int, java.util.Iterator)} specifically for jsonpath look-ups.
+     */
+    @CheckForNull
+    public KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, Iterator<JsonPathSegmentReference> jsonPath) {
```
gavlyukovskiy, line 181: I think it would be better to simply call the `getMaskConfigIfMatched(bytes, 0, 0, jsonPath)` from here. This basically ensures that return is called immediately

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -174,28 +164,101 @@ public byte[] mask(byte[] input) {
...
+            maskingState.backtrackCurrentJsonPath();
         }
         // Check if this is the start of a json array
         if (currentByteIsUnescapedSquareBracketOpen(maskingState)) {
             maskingState.expandCurrentJsonPath();
+            // Check is the first element in the array has to be masked.
+            KeyMaskingConfig keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), maskingState.getCurrentJsonPath());
+            if (keyMaskingConfig != null) {
+                maskArrayElementInPlace(maskingState, keyMaskingConfig);
+                trackCurrentJsonPath(maskingState);
```
gavlyukovskiy, line 241: Does it make sense to move it inside `maskArrayElementInPlace`?

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -174,28 +164,101 @@ public byte[] mask(byte[] input) {
...
+     * @param maskingState the masking state with the index on the element prior to the element start index
+     */
+    private void maskArrayElementInPlace(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
+        maskingState.incrementCurrentIndex();
+        skipWhitespaceCharacters(maskingState);
+        boolean maskable = maskValue(maskingState, keyMaskingConfig);
+        if (!maskable) {
+            // non-maskable values
+            skipAllValues(maskingState);
+        }
```
gavlyukovskiy, line 178: Is this only applied for `nulls` and when number / boolean masking is disabled? If I understand correctly for objects and arrays we always skip values inside `maskArrayValueInPlace` and `maskObjectValueInPlace`. I'd rather move this logic inside `mask*ValueInPlace`, otherwise it's slightly confusing

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: While looking more into it and this way of tracking makes it quite confusing - we have explicit expanding and backtracking in some places, but implicit inside `trackCurrentJsonPath` that does it based on the current state. Isn't this particular occurrence is simply `maskingState.incrementCurrentJsonPathArrayIndex();`?

donavdey: > Isn't this particular occurrence is simply maskingState.backtrackCurrentJsonPath();

No really. It could be expanding to a new array segment (next array element is an array), incrementing the index in the current array segment, backtracking current array segment or exapnding to a new object segment (next array element is an object).

> Does it make sense to move it inside maskArrayElementInPlace?

I kept it in the track method so that "maskArrayElementInPlace" is independent of the tracking logic.
In general, I moved as much tracking logic as possible inside "trackCurrentJsonPath" method. This way it is easier to disable it if there are no jsonpath keys.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: `maskValue` would return false only for null values. I created this method because sonarqube failed on code duplication as I used the same sequence of "if-else" statements to check the value type as in `maskObjectValueInPlace` method. The methods that use `maskValue` handle non-maskable values differently, so I cannot move the handling to `maskValue` method.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: -1? 😅

donavdey: I thought about it too! But it is "index out of bound" for "knownByteLengths" array. i could have made a special check for that, but decided that creating an overload is clearer.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -207,10 +233,17 @@ private TrieNode searchForJsonPathKeyNode(byte[] bytes, Iterator<MaskingState.Se
...
             }
-            MaskingState.SegmentReference segmentReference = jsonPath.next();
-            int keyStartIndex = segmentReference.start;
-            int keyLength = segmentReference.offset;
-            for (int i = keyStartIndex; i < keyStartIndex + keyLength; i++) {
+            JsonPathSegmentReference jsonPathSegmentReference = jsonPath.next();
+            int keyOffset = jsonPathSegmentReference.getOffset();
+            int keyStart = jsonPathSegmentReference.getLength();
+            if (jsonPathSegmentReference.isArraySegment()) {
+                node = node.children[Character.forDigit(keyOffset, DECIMAL_RADIX) + BYTE_OFFSET];
```
donavdey, line 240: I just realised that it won't work when the array index is > 10. I will push a fix and tests that cover such use cases.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -207,10 +233,17 @@ private TrieNode searchForJsonPathKeyNode(byte[] bytes, Iterator<MaskingState.Se
             if (node == null) {
                 return null;
             }
-            MaskingState.SegmentReference segmentReference = jsonPath.next();
-            int keyStartIndex = segmentReference.start;
-            int keyLength = segmentReference.offset;
-            for (int i = keyStartIndex; i < keyStartIndex + keyLength; i++) {
+            JsonPathSegmentReference jsonPathSegmentReference = jsonPath.next();
+            int keyOffset = jsonPathSegmentReference.getOffset();
+            int keyStart = jsonPathSegmentReference.getLength();
```
gavlyukovskiy, line 238: Isn't it `keyLength`?

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: Strange that we didn't catch that in our fuzzing tests

gavlyukovskiy: Ah 😅 
image

@Breus I think we should use at least size 10 for all generation options

Breus: Lol

Breus: Maybe also increase some other values to be sure?

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: I have tried with additional check and I think it would be slightly better, mostly because `getMaskConfigIfMatched` contains quite a lot of comments on why it does things in certain way, e.g.
image

WDYT about [this](https://github.com/Breus/json-masker/files/14421650/Use_-1_for_key_length.patch)?

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: Okay, I thought it was more straightforward

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: Wouldn't it make sense to move `skipAllValues` inside `maskValue` then? I've tried [this](https://github.com/Breus/json-masker/files/14421714/maskValue_refactor.patch) and all tests seem to pass.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: Yes, I think we need to do it for all of them, so it's at least 10

Breus: @donavdey can you update them in your PR?

gavlyukovskiy: @Breus I've tried increasing all of them to 12, but none of the tests failed 🤷  Apparently Java decided that it's better to return a null character `\u0000` if input is incorrect and then our trie simply doesn't have null characters to match. We'll need a specific test to cover that

gavlyukovskiy: With this
```java
                char digit = Character.forDigit(keyOffset, DECIMAL_RADIX);
                if (digit == '\u0000') {
                    throw new IllegalStateException("Invalid digit " + keyOffset);
                }
                node = node.children[digit + BYTE_OFFSET];
```
we actually have fuzzing test failure

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
donavdey: I think you are right. This way we don't have to duplicate the logic and have only one place where we would need to update it if anything.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: If the tests pass then let's contain the skipping logic inside `maskValue`.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: Committed the fix

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: Done

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: Added a check for null character

src/main/java/dev/blaauwendraad/masker/json/JsonPathSegmentReference.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * A mutable reference to a sequence of bytes in <code>dev.blaauwendraad.masker.json.MaskingState#message</code>. It is used to represent json path segments.
+ * <p>
+ * A reference is represented by a pair of integers:
+ * <ul>
+ *  <li><code>offset</code> denotes the start index of a segment in the message. In case the segment is an array, <code>offset</code> denotes the element index.</li>
+ *  <li><code>length</code> denotes the length of a segment in the message. In case the segment is an array, <code>length</code> is set to -1.</li>
+ * </ul>
+ */
+public class JsonPathSegmentReference {
```
gavlyukovskiy, line 12: Maybe that's a good candidate for sealed interface: `ArrayPathReference(int index)` and `NodePathReference(offset, length)`?

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
donavdey: It is. I forgot to rename variable name after I renamed the fields.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -207,15 +219,39 @@ private TrieNode searchForJsonPathKeyNode(byte[] bytes, Iterator<MaskingState.Se
...
+                for (int i = keyOffset; i < keyOffset + keyLength; i++) {
+                    int b = bytes[i];
+                    node = node.children[b + BYTE_OFFSET];
+                    if (node == null) {
+                        return null;
+                    }
                 }
+            } else if (jsonPathSegmentReference instanceof JsonPathSegmentReference.Array jsonPathArray) {
+                // for arrays keyOffset is the index of the element
+                if (jsonPathArray.getIndex() > 10) {
```
gavlyukovskiy, line 235:
```suggestion
                if (jsonPathArray.getIndex() >= 10) {
```

src/main/java/dev/blaauwendraad/masker/json/JsonPathSegmentReference.java
gavlyukovskiy: @donavdey I've pushed a commit making it sealed, let me know if it looks worse to you and we can revert it

gavlyukovskiy:
@donavdey I'm done messing with your PR :)

gavlyukovskiy (approved):

src/main/java/dev/blaauwendraad/masker/json/JsonPathSegmentReference.java
donavdey: This looks good. Now I finally see how sealed interfaces are useful in the first place.

PR #63, opened 2024-02-27
Improve fuzzing tests

gavlyukovskiy:
Make it run longer on CI, increase all limits for generated JSONs.


Breus (approved):
Nice changes :-)

PR #73, opened 2024-03-03
Refactor main loop and masking in place with visitors

gavlyukovskiy:
This change simplifies the masking by using visitors, whenever we encounter an unknown value (including the full json) we call `visitValue` that decide how to process it further:
1. In allow mode we always start with default masking config as everything should be masked by default. In mask mode we start with `null` (no masking)
2. Primitive values are masked if the masking config is not null
3. Array type has its own visitor method `visitArray`, which calls `visitValue` on each element. They always reuse the config supplied with no additional lookup (which is the reason I didn't want to restore support for indexed json paths)
4. Object type has its own visitor method `visitObject`, which iterates over the keys, checks whether key has a masking config (in allow mode whether it's explicitly allowed) and calls `visitValue` with the config (if key needs to be masked) or with null (if key does not need to be masked)

The flow can be visualized as following:
```mermaid
flowchart TD
    mask -->|"visitValue(allowMode ? defaultConfig : null)"| visitValue
    visitValue -->|is primitive| visitPrimitive
    visitValue -->|is object| visitObject
    visitValue -->|is array| visitArray
    visitArray -->|each element| visitValue
    visitPrimitive -->|keyMaskingConfig != null| maskValue
    visitPrimitive -->|keyMaskingConfig == null| skipValue
    visitObject -->|each key| lookupKeyMaskingConfig 
    lookupKeyMaskingConfig -->|"explicitly allowed (in allow mode)"| skipValue
    lookupKeyMaskingConfig -->|each value| visitValue
```

Additional changes:
- Adds support of masking top level primitive values and arrays (only applicable to allow mode)
- Disables support for JsonPath index masking as we're not going to support it (and also to simplify lookup).
- Refactored unformatted json tests to not use files as it's harder to manage


donavdey (approved):
The change looks very good. Technically we can keep this visitor solution instead of implementing a tokenizer / jsonpath-based state machine as the masking process is clear enough now .

Breus (approved):

PR #75, opened 2024-03-03
Lower json size for jmh

### LINKED ISSUE #74: Update README to reflect better the typical JSON size
We currently use `2mb` JSON in the readme, which is a bit excessive, it would be better to update it to `1kb`.

Also makes sense to update GitHub action for benchmarking to show that for every PR

gavlyukovskiy:
- [x] Wait for #73 to be merged
- [x] Wait for #82 to check results on the smaller JSON size
- [x] Wait for benchmarks and update readme
- [x] ~~Add comparison to byte version of jackson~~ (byte version of jackson is comparable to string version)

Fixes #74


build.gradle.kts
```diff
@@ -124,13 +124,13 @@ jmh {
     // run this with ./gradlew jmh -PjmhShort to only run these parameters and 4 iterations
     if (project.hasProperty("jmhShort")) {
         benchmarkParameters = mapOf(
-            "jsonSize" to listOfProperty("2mb"),
-            "maskedKeyProbability" to listOfProperty("0.01"),
+            "jsonSize" to listOfProperty("1kb"),
+            "maskedKeyProbability" to listOfProperty("0.1"),
             "jsonPath" to listOfProperty("false", "true"),
             "characters" to listOfProperty("unicode")
```
gavlyukovskiy, line 130:
```suggestion
             "characters" to listOfProperty("ascii")
```

Breus (approved):

PR #78, opened 2024-03-04
Replace RandomWhiteSpaceInjector by PrettyPrinter impl

Sprokof:


gavlyukovskiy:
@Sprokof thank you for the PR, but it looks like it misses crucial context on what the `PrettyPrinter` implementation supposed to do - we want to inject random amount of whitespaces when printing a jackson, not actually "pretty print" it. We have another PR opened https://github.com/Breus/json-masker/pull/79 that (almost) does what we want it to do, so I'll be closing this PR and continue in #79. 
If you want to help, feel free to review that PR or suggest some other improvements 👍

PR #79, opened 2024-03-04
Replace RandomWhiteSpaceInjector to Jackson PrettyPrinter

AlexeyShary:
Changed RandomJsonWhiteSpaceInjector to work with RandomWhiteSpacePrettyPrinter.
RandomWhiteSpacePrettyPrinter is a DefaultPrettyPrinter with a logic to add white spaces.
But now it works slower.

### DIFF
--- a/src/test/java/dev/blaauwendraad/masker/randomgen/RandomWhiteSpacePrettyPrinter.java
+++ b/src/test/java/dev/blaauwendraad/masker/randomgen/RandomWhiteSpacePrettyPrinter.java
@@ -1,0 +1,104 @@
+package dev.blaauwendraad.masker.randomgen;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.PrettyPrinter;
+import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
+
+import java.io.IOException;
+import java.util.concurrent.ThreadLocalRandom;
+
+import static dev.blaauwendraad.masker.json.util.AsciiCharacter.*;
+
+/**
+ * This class is DefaultPrettyPrinter with additional logic to add white spaces to JSON
+ */
+public class RandomWhiteSpacePrettyPrinter implements PrettyPrinter {
+    private final DefaultPrettyPrinter defaultPrettyPrinter = new DefaultPrettyPrinter();
+
+    private int numberOfWhiteSpacesToInject;
+
+    public RandomWhiteSpacePrettyPrinter(int numberOfWhiteSpacesToInject) {
+        this.numberOfWhiteSpacesToInject = numberOfWhiteSpacesToInject;
+    }
+
+    @Override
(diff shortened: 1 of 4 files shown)


src/test/java/dev/blaauwendraad/masker/randomgen/RandomWhiteSpacePrettyPrinter.java
```diff
@@ -0,0 +1,104 @@
...
+            gen.writeRaw((char) getRandomWhiteSpaceByte(randomInt));
+            numberOfWhiteSpacesToInject--;
+        }
+    }
+
+    private byte getRandomWhiteSpaceByte(int randomInt) {
+        return switch (randomInt) {
+            case 0 -> HORIZONTAL_TAB.getAsciiByteValue();
+            case 1 -> LINE_FEED.getAsciiByteValue();
+            case 2 -> SPACE.getAsciiByteValue();
```
gavlyukovskiy, line 100: Can you also include `CARRIAGE_RETURN` here? I've accidentally removed it in #71 in attempt to fix the problem, but it wasn't enough.

src/test/java/dev/blaauwendraad/masker/randomgen/RandomWhiteSpacePrettyPrinter.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public void beforeArrayValues(JsonGenerator gen) throws IOException {
+        defaultPrettyPrinter.beforeArrayValues(gen);
+        addRandomWhiteSpace(gen);
+    }
+
+    @Override
+    public void beforeObjectEntries(JsonGenerator gen) throws IOException {
+        defaultPrettyPrinter.beforeObjectEntries(gen);
+        addRandomWhiteSpace(gen);
```
gavlyukovskiy, line 80: Does it make sense to inject the whitespace not only after, but also before (here and in other places)?

AlexeyShary:
@gavlyukovskiy thank you for the review! It seems now I've grasped the idea. Made changes according to all the comments:

- NoFailingExecutionFuzzingTest.executeTest generates a JsonNode for further processing
- The RandomJsonWhiteSpaceInjector class is removed, and its logic is moved into JsonFormatter
- The signature of NoFailingExecutionFuzzingTest.executeTest jsonMutator is changed from <byte[], byte[]> to <JsonNode, String> - generating JsonNode during the test and converting it into a string via JsonFormatter, which uses RandomWhiteSpacePrettyPrinter
- Included CARRIAGE_RETURN in the generated characters
- Added generation of spaces before and after the output characters for greater variation in results

Now it works much faster.

gavlyukovskiy:
@AlexeyShary thank you for the PR, it looks good now. Don't mind the errors, I think we have some problems with the pipeline running on a forked repositories due to usage of secrets, we'll need to fix those before merging

PR #84, opened 2024-03-04
Try going full r* mode and use naked index instead of state

gavlyukovskiy:
Slightly inspired by https://github.com/alibaba/fastjson2/blob/main/core/src/main/java/com/alibaba/fastjson2/JSONReader.java I wonder if that's significantly faster than what we have now


gavlyukovskiy:
It is not 😅

gavlyukovskiy:
Let's check on the smaller jsons

gavlyukovskiy:
About 30% difference, not bad, but I'm not looking forward to using indexes again 😓

PR #87, opened 2024-03-06
Improve baseline implementation for benchmarks

donavdey:
The PR improves the baseline benchmark performance by eliminating jsonpath keys parsing into internal `JsonPath` objects.  
The introduction of `dev.blaauwendraad.masker.json.path.JsonPathParser#isNumber` previously decreased the baseline benchmark performance by 4 times. Let's see by how much the performance will increase now.


donavdey:
Basically it restored back the original performance.

gavlyukovskiy (approved):

PR #92, opened 2024-03-11
Fix ArrayIndexOutOfBoundsException when skipping allowed string value with escaped escape character

gavlyukovskiy:
This PR includes some refactoring for the string masking in general, notably that the string is masked including the quotes (internal change, from the API the quotes were not required and that hasn't changed)


Breus:
@gavlyukovskiy how did you stumble upon this bug?

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -434,14 +393,52 @@ private static void skipArrayValue(MaskingState maskingState) {
...
-        return AsciiCharacter.isDoubleQuote(maskingState.byteAtCurrentIndex())
-                && !AsciiCharacter.isEscapeCharacter(maskingState.byteAtCurrentIndexMinusOne());
+    private static int expandStringValue(MaskingState maskingState) {
+        maskingState.incrementCurrentIndex(); // step over the JSON key opening quote
+        int nonVisibleCharacters = 0;
+        boolean isEscapeCharacter = false;
+        boolean previousCharacterCountedAsEscapeCharacter = false;
+        while (!AsciiCharacter.isDoubleQuote(maskingState.byteAtCurrentIndex())
+               || (AsciiCharacter.isDoubleQuote(maskingState.byteAtCurrentIndex()) && isEscapeCharacter)) {
+            if (Utf8Util.getCodePointByteLength(maskingState.byteAtCurrentIndex()) > 1) {
```
gavlyukovskiy, line 410: Ideally we should not count the non-visible characters for every string expansion, it only makes sense when we mask string characters, not the whole string. Not sure how much is the performance impact

gavlyukovskiy: Did that in the latest commit, now we skip string without counting the non-visible characters and only count if when we have specific configuration

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -141,21 +140,18 @@ private void visitObject(MaskingState maskingState, @CheckForNull KeyMaskingConf
...
             // In case target keys should be considered as allow list, we need to NOT mask certain keys
             int openingQuoteIndex = maskingState.currentIndex();
-            maskingState.incrementCurrentIndex(); // step over the JSON key opening quote
-            while (!currentByteIsUnescapedDoubleQuote(maskingState)) {
-                maskingState.incrementCurrentIndex();
-            }
+
+            skipStringValue(maskingState);
 
             int closingQuoteIndex = maskingState.currentIndex();
```
Breus, line 146: At this point the current index is actually one step AFTER the closing quote, not on the closing quote, so this variable name is now wrong.

Breus: You could:

`int afterClosingQuoteIndex = maskingState.currentIndex()` and keep -2 for the quotes you do later

or 

`int closingQuoteIndex = maskingState.currentIndex() - 1` and do minus 1 quote later

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -197,50 +193,12 @@ private void visitObject(MaskingState maskingState, @CheckForNull KeyMaskingConf
...
-                    additionalBytesForEncoding += 4;
-                }
-                previousCharacterCountedAsEscapeCharacter = false;
-            }
-            targetValueLength++;
-            maskingState.incrementCurrentIndex();
-        }
+        int valueStartIndex = maskingState.currentIndex();
+        skipStringValue(maskingState);
+        int targetValueLength = maskingState.currentIndex() - valueStartIndex;
```
Breus, line 198: This definitely requires one or two lines of documentation 😄 
It is not super obvious to see why you jump over the String to mask it

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -197,50 +193,12 @@ private void visitObject(MaskingState maskingState, @CheckForNull KeyMaskingConf
      * @param keyMaskingConfig the {@link KeyMaskingConfig} for the corresponding JSON key
```
Breus, line 193: Would be nice to add to this method JavaDoc where the index will be after masking

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Actually, now that I think of it. I think we should rename all `skip[Type]Value` methods with `stepOver[Type]Value` that would make it clear for both cases right away what is happening. WDYT?

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -316,17 +273,17 @@ private void maskNumber(MaskingState maskingState, KeyMaskingConfig keyMaskingCo
      */
     private void maskBoolean(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
         int targetValueLength = AsciiCharacter.isLowercaseT(maskingState.byteAtCurrentIndex()) ? 4 : 5;
```
Breus, line 275: I will add a comment that this is 4 because of `true` or 5 because of either `null` or `false`

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: I think stepOver sounds good

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: Here it's only true (4) or false (5). Did you mean visitValue method where we do `'t', 'n' -> 4`?

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: Good catch :+1:

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: Good catch 👍 Let's do the first suggestion. Or maybe we could also return the length of the key from skipStringValue method? Though we still need to remember the index before

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Ah, you are right

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Eh, this is fine for now. I will send an update commit later

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Makes sense!

Breus (approved):
I've added a commit that addresses my previous review points

PR #95, opened 2024-03-15
Create ADR for jsonpath support

donavdey:


adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Status
+
+Draft
+
+## Context
+
+Support for JsonPATH was requested by some potential library clients.  
+However, JsonPATH RFC specifies way too many features which are tedious to implement without having runtime dependencies.  
+Therefore, we have to decide which features are necessary and which are not.
```
Breus, line 13: The reason we don't support all JsonPath features is actually not because it is tedious to implement, but there are three other main reasons:

- First of all, JsonPath support was added to be able to map Java class fields to JsonPaths because if you would have the same field nested in a different field it is impossible to distinguish the two without JsonPath. The limited JsonPath we currently support is precisely what is required to fully support this use case (e.g. works both with collection fields as with nested objects).
- Second, with every feature we add to the library, we add maintenance complexity and the probably to introduce bugs. Since no-one requested any of the other JsonPath features, there is no point to add it risking us to introduce a bug besides the currently supported JsonPath features.
- Finally, with every feature we add to the library performance usually gets slightly worse which again, is not worth it considering no-one ever requested any other JsonPath feature besides the ones you implemented so far :-)

Breus: Tedious to implement is basically not a reason, because if that would be the case we could just open an issue and ask the community to implement it or do it ourselves on a rainy Sunday :-)

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+## Context
+
+Support for JsonPATH was requested by some potential library clients.  
+However, JsonPATH RFC specifies way too many features which are tedious to implement without having runtime dependencies.  
+Therefore, we have to decide which features are necessary and which are not. 
+
+## Decisions
+
+The decisions are premised on the following expectations:
+1. Most of the clients will not use JsonPATH keys.
```
Breus, line 18: And more importantly: not require JsonPath support besides using it to target unique fields corresponding to (annotated) Java fields.

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+Support for JsonPATH was requested by some potential library clients.  
+However, JsonPATH RFC specifies way too many features which are tedious to implement without having runtime dependencies.  
+Therefore, we have to decide which features are necessary and which are not. 
+
+## Decisions
+
+The decisions are premised on the following expectations:
+1. Most of the clients will not use JsonPATH keys.  
+2. The correctness of the library takes precedent over the amount of features.
```
Breus, line 19: correctness, performance and maintainability always come before any non-requested feature basically

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+Any implemented jsonpath feature adds some performance overhead both on configuration and on masking time, even if it is not used. In general, it also increases the likelihood of incorrect behaviour as the amount of code increases.
```
Breus, line 21: Nit: write `jsonpath` consistently (in whatever way you see fit)

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+Therefore, we decided to keep support for JsonPATH as limited as possible and to extend it only when someone specifically requests it.
```
Breus, line 22: The aforementioned reasons are all correct. However, the main reason we choose for the current set of JsonPATH features is to support masking annotated fields in a Java object which could potentially have duplicate names in nested objects

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+The following features are deemed necessary:
+1. Support for bracket, dot and mixed notations.
```
Breus, line 25: Bracket, dot and mixed notations is supported just for flexibility of our API as it was (as you said) relatively simple to implement without adding tons of complexity or performance overhead.

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+Any implemented jsonpath feature adds some performance overhead both on configuration and on masking time, even if it is not used. In general, it also increases the likelihood of incorrect behaviour as the amount of code increases.
+Therefore, we decided to keep support for JsonPATH as limited as possible and to extend it only when someone specifically requests it.
+
+The following features are deemed necessary:
+1. Support for bracket, dot and mixed notations.  
+2. Support for object segments.  
+3. Support for array segments with wildcard indexing only.  
+4. Wildcard segments and selectors.   
+
+The rest of the features are deemed unnecessary.
```
Breus, line 30: for now.

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Consequences
+
+### Performance considerations.
+
+Supporting JsonPATH makes the masking 25% slower according to the latest benchmarks.  
+We decided to disable JsonPATH in case no JsonPATH keys are supplied.
```
Breus, line 37: Therefore, we decided

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Potential bugs
+
+Mixing keys and JsonPATH keys in the same trie opens up some bugs:
+1. https://github.com/Breus/json-masker/issues/94  
+We do not expect this issue to be reproduced naturally.
```
Breus, line 43: I wouldn't add this for two reasons:
- As you mention, this is ridiculously unlikely to appear in the wild
- This is just something fixable and doesn't come directly as a consequence of supporting JsonPath but just because of some implementation detail that we decide about.

adr/0003-jsonpath-support.md
gavlyukovskiy: https://www.rfc-editor.org/rfc/rfc9535.html RFC specifies `JSONPath`

adr/0003-jsonpath-support.md
gavlyukovskiy: Also I would add that JSONPath was made first and foremost for selecting values, but some things simply don't make sense for masking (reasons for ambiguity checks for example)

adr/0003-jsonpath-support.md
gavlyukovskiy, line 30: I would rather document the features we support according to [the RFC](https://www.rfc-editor.org/rfc/rfc9535.html), for example
1. [Name Selector](https://www.rfc-editor.org/rfc/rfc9535.html#name-name-selector) - Full support
2. [Wildcard Selector](https://www.rfc-editor.org/rfc/rfc9535.html#name-wildcard-selector) - Full support, except ambiguous segments
3. [Index Selector](https://www.rfc-editor.org/rfc/rfc9535.html#name-index-selector) - Not supported
4. [Array Slice Selector](https://www.rfc-editor.org/rfc/rfc9535.html#name-array-slice-selector) - Not supported
5. [Filter Selector](https://www.rfc-editor.org/rfc/rfc9535.html#name-filter-selector) - Not supported

donavdey:
@Breus @gavlyukovskiy I elaborated the ADR according to the comments.

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,78 @@
...
+JsonPATH RFC specifies a wide variety of features. Not all of them are required to solve the described problem.  
+Therefore, we have to decide which features are necessary and which are not. 
+  
+The decision is premised on the following expectations:
+1. Most of the clients will not face the name ambiguity problem, therefore they will not use JsonPATH.  
+2. Correctness, performance and maintainability of the library takes precedent over the amount of features.  
+3. JsonPATH support is required only for solving the name ambiguity problem.  
+
+Any implemented JsonPATH feature adds some performance overhead both on configuration and on masking time, even if it is not used. In general, it also increases the likelihood of incorrect behaviour as the amount of code increases. Therefore, having unnecessary features violates the premises.   
+Therefore, we decided to keep support for JsonPATH as limited as possible and extend it only when someone specifically requests it.
```
gavlyukovskiy, line 46: Therefore the same as in previous sentence, maybe change that to something else?

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,78 @@
...
+We expect developers to know either of these notations, but we cannot know which one exactly.  
+2. **Support for segments**.  
+This is the main building block of JsonPATH queries.
+3. **Support for wildcard segments and selectors**.  
+Wildcard selectors are necessary for traversing array values. 
+
+## Not supported features
+
+1. **Name selectors**.  
+Using segments is enough to select any value in the json.
```
gavlyukovskiy, line 59: I thought we fully support name selectors, if I understand it correctly, in the spec "segment" is used to describe any segment, including name selector  which is the simplest selector (`$.a` is a name selector)

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Supported features  
+
+JsonPATH RFC specifies a wide variety of features. Not all of them are required to solve the described problem.  
+Therefore, we have to decide which features are necessary and which are not. 
+  
+The decision is premised on the following expectations:
+1. Most of the clients will not face the name ambiguity problem, therefore they will not use JsonPATH.  
+2. Correctness, performance and maintainability of the library takes precedent over the amount of features.  
+3. JsonPATH support is required only for solving the name ambiguity problem.
```
gavlyukovskiy, line 43: Nice summary!

adr/0003-jsonpath-support.md
donavdey: I think you are right. Let me fix that.

adr/0003-jsonpath-support.md
donavdey: Good one, I will change it.

Breus (approved):

adr/0003-jsonpath-support.md
```diff
@@ -0,0 +1,93 @@
...
+
+The solution to the described problem is to provide a way to disambiguate key/value pairs by selecting the target pair
+using its location in JSON.  
+The industry standard for selecting values in JSON is JsonPATH. Most developers are expected to know how to create basic
+JsonPATH queries using either bracket or dot notation.  
+Therefore, we decided to solve the key ambiguity problem using JsonPATH.
+
+### Supported features
+
+The JsonPATH RFC specifies a wide variety of features. Not all of them are required to solve the described problem.
```
gavlyukovskiy, line 47:
```suggestion
The [JsonPATH RFC 9535](https://www.rfc-editor.org/rfc/rfc9535.html) specifies a wide variety of features. Not all of them are required to solve the described problem.  
```
Let's add a link to RFC

gavlyukovskiy (approved):

PR #96, opened 2024-03-15
Refactor all value masking through a unified interface

gavlyukovskiy:
Practical implementation for https://github.com/Breus/json-masker/pull/89#pullrequestreview-1926179868

TODO:
- [ ] Docs for all new functionality
- [ ] Performance to be comparable when using the `ValueMasker`
- [ ] Review naming of `ValueMasker` / `ValueMaskerContext`
- [ ] Decide on which maskers should be provided out of the box (this PR adds `ValueMasker.maskEmail` and `ValueMasker.maskStringWithFunction`)

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
@@ -4,7 +4,6 @@ import dev.blaauwendraad.masker.json.config.JsonMaskingConfig;
 import dev.blaauwendraad.masker.json.config.KeyMaskingConfig;
 import dev.blaauwendraad.masker.json.util.AsciiCharacter;
 import dev.blaauwendraad.masker.json.util.AsciiJsonUtil;
-import dev.blaauwendraad.masker.json.util.Utf8Util;
 
 import javax.annotation.CheckForNull;
 import java.util.Collections;
@@ -73,7 +72,7 @@ public final class KeyContainsMasker implements JsonMasker {
             case '[' -> visitArray(maskingState, keyMaskingConfig);
             case '{' -> visitObject(maskingState, keyMaskingConfig);
             case '-', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' -> {
-                if (keyMaskingConfig != null && !keyMaskingConfig.isDisableNumberMasking()) {
+                if (keyMaskingConfig != null) {
                     maskNumber(maskingState, keyMaskingConfig);
                 } else {
                     stepOverNumericValue(maskingState);
@@ -87,20 +86,20 @@ public final class KeyContainsMasker implements JsonMasker {
                 }
             }
             case 't' -> {
-                if (keyMaskingConfig != null && !keyMaskingConfig.isDisableBooleanMasking()) {
+                if (keyMaskingConfig != null) {
                     maskBoolean(maskingState, keyMaskingConfig);
                 } else {
-                    maskingState.setCurrentIndex(maskingState.currentIndex() + 4);
+                    maskingState.incrementCurrentIndex(4);
                 }
             }
             case 'f' -> {
-                if (keyMaskingConfig != null && !keyMaskingConfig.isDisableBooleanMasking()) {
+                if (keyMaskingConfig != null) {
                     maskBoolean(maskingState, keyMaskingConfig);
                 } else {
-                    maskingState.setCurrentIndex(maskingState.currentIndex() + 5);
+                    maskingState.incrementCurrentIndex(5);
                 }
             }
-            case 'n' -> maskingState.setCurrentIndex(maskingState.currentIndex() + 4);
+            case 'n' -> maskingState.incrementCurrentIndex(4);
         }
     }
 
@@ -201,42 +200,12 @@ public final class KeyContainsMasker implements JsonMasker {
      * @param keyMaskingConfig the {@link KeyMaskingConfig} for the corresponding JSON key
      */
     private void maskString(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
-        int valueStartIndex = maskingState.currentIndex();
-        stepOverStringValue(maskingState); // we step over the string to find its length to mask it
-        int targetValueLength = maskingState.currentIndex() - valueStartIndex;
-        if (keyMaskingConfig.getMaskStringsWith() != null) {
-            maskingState.replaceTargetValueWith(
-                    valueStartIndex,
-                    targetValueLength,
-                    keyMaskingConfig.getMaskStringsWith(),
-                    1
-            );
-        } else if (keyMaskingConfig.getMaskStringCharactersWith() != null) {
-            /*
-            So we don't add asterisks for escape characters or additional encoding bytes (which are not part of the String length)
+        maskingState.setCurrentValueStartIndex();
+        stepOverStringValue(maskingState);
 
-            The actual length of the string is the length minus escape characters (which are not part of the
-            string length). Also, unicode characters are denoted as 4-hex digits but represent actually
-            just one character, so for each of them 3 asterisks should be removed.
-             */
-            valueStartIndex += 1; // offset by opening quote
-            targetValueLength -= 2; // remove quotes from the value length
+        keyMaskingConfig.getStringValueMasker().maskValue(maskingState);
 
-            int nonVisibleCharacters = Utf8Util.countNonVisibleCharacters(
-                    maskingState.getMessage(),
-                    valueStartIndex,
-                    targetValueLength
-            );
-
-            maskingState.replaceTargetValueWith(
-                    valueStartIndex,
-                    targetValueLength,
-                    keyMaskingConfig.getMaskStringCharactersWith(),
-                    targetValueLength - nonVisibleCharacters
-            );
-        } else {
-            throw new IllegalStateException("Invalid string masking configuration");
-        }
+        maskingState.unsetCurrentValueStartIndex();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -309,9 +257,9 @@ private void maskBoolean(MaskingState maskingState, KeyMaskingConfig keyMaskingC
     private static void stepOverValue(MaskingState maskingState) {
         switch (maskingState.byteAtCurrentIndex()) {
             case '"' -> stepOverStringValue(maskingState);
-            case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '-' -> stepOverNumericValue(maskingState);
-            case 't', 'n' -> maskingState.setCurrentIndex(maskingState.currentIndex() + 4); // true or null
-            case 'f' -> maskingState.setCurrentIndex(maskingState.currentIndex() + 5); // false
+            case '-', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' -> stepOverNumericValue(maskingState);
```
gavlyukovskiy, line 260: Keeping OCD [in check](https://github.com/Breus/json-masker/pull/96/files#diff-e8659d668c0ffdcb0d5d6fa250b0d23ad7e2947f9d5eb2d8575b4e2b04065179R74) 😅

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -252,26 +221,12 @@ The actual length of the string is the length minus escape characters (which are
      */
     private void maskNumber(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
         // This block deals with numeric values
-        int targetValueStartIndex = maskingState.currentIndex();
+        maskingState.setCurrentValueStartIndex();
```
gavlyukovskiy, line 224: I have to set index of the current value start / end in order to calculate correct offset in `ValueMaskingContext`, if there's any ideas how to make it better, I would be happy to implement those

src/main/java/dev/blaauwendraad/masker/json/config/KeyMaskingConfig.java
```diff
@@ -158,130 +74,131 @@ private Builder() {
...
          */
         public Builder disableNumberMasking() {
-            if (disableNumberMasking != null) {
-                throw new IllegalArgumentException("'disableNumberMasking()' was already set");
-            }
-            checkMutuallyExclusiveNumberMaskingOptions();
-            disableNumberMasking = true;
+            maskNumbersWith(ValueMasker.noop());
             return this;
         }
```
gavlyukovskiy, line 126: I'm thinking of getting rid of `disableNumberMasking` / `disableBooleanMasking`, this now can be achieved with `ValueMasker.noop()` and it's not significant enough to have a separate API for that

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
Breus: Just for the record: this means that the index will be set to the start index of the current value? If so, than the following name might be slightly more clear:    `setIndexToValueStartIndex`

src/main/java/dev/blaauwendraad/masker/json/config/KeyMaskingConfig.java
Breus: Makes total sense to me.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: I would thinking of also making it less implementation specific and have something like `maskingState.beforeMaskableValue()` / `maskingState.afterMaskableValue()`

donavdey: Binding the name to the implementation details raises some issues. When we frame it as if we set the index of the current value to be masked, then I would expect that we do it for every value type, including array and object values. But this is not the case currently.
So I agree with Artur on making it less implementation specific. One option would be to name it `registerIndex` / `clearRegisteredIndex`

src/main/java/dev/blaauwendraad/masker/json/ValueMasker.java
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * Masks a target value with a supplied {@link Function}. The target value is passed into the function as a string,
+     * regardless of the type (string, numeric or a boolean), however any non-null returned value from the
+     * function will always be a JSON string (with quotes added automatically).
+     *
+     * <p>Note: usually {@link ValueMasker} operates on a byte level without parsing JSON values into intermediate
+     * objects, however this implementation will have to allocate a {@link String} before passing it into the function
+     * and then turn it back into a byte array for replacement.
+     */
+    static ValueMasker maskWithStringFunction(Function<String, String> masker) {
```
donavdey, line 150: Based purely on API client's perspective, it would be more intuitive if I can just pass a lambda `maskNumber((value) -> "888"`, then for presets I do `maskString(ValueMasker::maskEmail)`. Then, on the builder layer we can decorate these lambdas with DescriptiveMakser. Also, `MaskingStrategy` would look more intuitive than `ValueMakser`.

donavdey: But I think the current API is also intuitive enough.

donavdey (approved):
The change looks very good. I left a couple of comments from API clients POV and also an opinion about the namings.

src/main/java/dev/blaauwendraad/masker/json/ValueMasker.java
gavlyukovskiy: > it would be more intuitive if I can just pass a lambda maskNumber((value) -> "888"

I think that's a nice idea, we could add another builder method like this
```java
        public Builder maskStringsWith(Function<String, String> masker) {
            maskStringsWith(ValueMasker.maskWithStringFunction(masker));
            return this;
        }
```
but the only problem is that currently `Function<String, String>` behaves more like "string representation of the value", which also applied to numeric types. We could add more of these, for example
```java
public Builder maskNumbersWith(Function<Number, Number> value)
public Builder maskBooleansWith(Function<Boolean, Boolean> value)
```
but then we quickly run into problem of impossibility to convert type during masking, there can't be two methods like this:
```java
public Builder maskNumbersWith(Function<Number, Number> value)
public Builder maskNumbersWith(Function<Number, String> value)
```

donavdey: The problem still holds though regardless of whether we pass a lambda directly or decorate it through ValueMasker. It's just the first way that is more intuitive. 
As for the problem itself, I think it is unlikely that someone would need to partially mask numbers without converting the value to strings. Same for booleans.

gavlyukovskiy: > The problem still holds though regardless of whether we pass a lambda directly or decorate it through ValueMasker.

This is not entirely true - inside `ValueMasker` we actually differentiate by the type of the function, so in addition to `ValueMasker.maskWithStringFunction(Function<String, String>)` we could add `ValueMasker.maskWithNumberFunction(Function<Number, Number>)` and `ValueMasker.maskWithNumberStringFunction(Function<Number, String>)`

> As for the problem itself, I think it is unlikely that someone would need to partially mask numbers without converting the value to strings. Same for booleans.

WIth partial masking yes, but this is more about generic function for masking, as a user I'd rather expect `Function<Number, Number>` to be the default (though working with `Number` and not concrete implementation is also not very friendly)

gavlyukovskiy:
@donavdey @breus I've updated the PR quite significantly in order to make sure `ValueMasker` is tied to a concrete JSON type. I was writing some tests and noticed that due to making `ValueMasker` generic, we also now can have a misuses of the API. For example before this change, we could do
```java
// "secret" -> "******"
builder.maskStringCharactersWith("*")
```
and the implementation behind was only applicable to string, but since it was extracted into `ValueMaskers.eachCharacterWith()`, which returns a generic `ValueMasker`, now it can be potentially applied to numbers or booleans:
```
builder.maskNumbersWith(ValueMaskers.eachCharacterWith("*"))
builder.maskBooleansWith(ValueMaskers.eachCharacterWith("*"))
```
which gives unexpected results - `123` is masked as `1*3` and `true` is masked as `t**e`, because the implementation assumes that first and last bytes are string quotes.

To make sure that implementation is always bound to a specific type, I've refactored `ValueMasker` to be a sealed interface with per-type implementations (non-sealed, so they can be extended), but now it's explicit to the user that `ValueMaskers.eachCharacterWith` is an instance of `ValueMasker.StringMasker`, while `builder.maskNumbersWith` only only accepts `ValueMasker.NumberMasker`.
There is also a special type `ValueMasker.AnyValueMasker`, which extends all 3 types and is used the implementation does not care about the actual type, i.e. `ValueMasker.with("***")` or `ValueMasker.withNull()` can be applied to any value , hence is accepted by any builder method.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
gavlyukovskiy: @donavdey we applied your suggestion with `register` / `clear`

src/test/java/dev/blaauwendraad/masker/json/ValueMaskersTest.java
```diff
@@ -0,0 +1,206 @@
...
+    @Nonnull
+    private byte[] everyOtherCharacterMasked(ValueMaskerContext valueMaskerContext) {
+        byte[] maskResult = new byte[valueMaskerContext.byteLength()];
+        for (int i = 0; i < valueMaskerContext.byteLength(); i++) {
+            if (i % 2 == 0) {
+                maskResult[i] = valueMaskerContext.getByte(i);
+            } else {
+                maskResult[i] = '*';
+            }
+        }
```
gavlyukovskiy, line 203: This implementation is slightly wrong - the string value must have been used with quotes (see other tests) and the implementation itself should have skipped the quotes, otherwise you might mask the closing quote with `*` producing an invalid JSON.

Breus: Yikes, the fact that I didn't remember that one day after discussing is not a good sign

Breus: I also was planning to write more end to end json masker like this instead of this unit actually. Should've done that...

PR #97, opened 2024-03-19
Fix cases when JsonMasker can get in infinite loop

gavlyukovskiy:
Fixes issue when we cannot find the end of an array, other validation is mostly for failing fast with more predictable error message.
Added a formatter to produce an invalid json.

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
@@ -53,6 +53,8 @@ import java.util.Collections;
                     Collections.emptyIterator()
             );
         }
+
+        stepOverWhitespaceCharacters(maskingState);
         visitValue(maskingState, keyMaskingConfig);
 
         return maskingState.flushReplacementOperations();
@@ -66,7 +68,6 @@ import java.util.Collections;
      *                         being masked
      */
     private void visitValue(MaskingState maskingState, @CheckForNull KeyMaskingConfig keyMaskingConfig) {
-        stepOverWhitespaceCharacters(maskingState);
         // using switch-case over 'if'-statements to improve performance by ~20% (measured in benchmarks)
         switch (maskingState.byteAtCurrentIndex()) {
             case '[' -> visitArray(maskingState, keyMaskingConfig);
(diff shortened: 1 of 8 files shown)


src/test/java/dev/blaauwendraad/masker/json/InvalidJsonTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class InvalidJsonTest {
```
gavlyukovskiy, line 15:
```suggestion
class InvalidJsonTest {
```

donavdey (approved):
I was not able to come up with examples that would stuck the masker. So I'd say it is safe enough.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -178,15 +190,15 @@ private void visitObject(MaskingState maskingState, @CheckForNull KeyMaskingConf
...
+            // check if we're at the end of a (non-empty) object
+            if (maskingState.byteAtCurrentIndex() == '}') {
+                break;
             }
-            stepOverWhitespaceCharacters(maskingState);
-
-            maskingState.backtrackCurrentJsonPath();
         }
-        maskingState.incrementCurrentIndex(); // step over closing curly bracket
+        maskingState.next();
```
Breus, line 201: Nit: it is nice to add the comment back (step over closing curly bracket) but if this is the only comment I can do it in my next PR

src/test/java/dev/blaauwendraad/masker/randomgen/InvalidJsonPrettyPrinter.java
```diff
@@ -0,0 +1,115 @@
...
+            '[',
+            ']',
+            ':',
+            '\'',
+            '"',
+            ',',
+            'a',
+            't',
+            'f',
+            'n'
```
Breus, line 34: Probably we should also add numeric characters (-, e, E, 0, 1) but can also do it in later PR

Breus (approved):

PR #103, opened 2024-03-27
Refactor withTextFunction to indicate that the value received is a raw JSON literal

gavlyukovskiy:
After finding some edge cases on the `withTextFunction`, that meant to return a textual representation of JSON value, we decided to change the function to always work on raw JSON literals, meaning that we also return the quotes for string values.

Couple of cases when the old behavior would be confusing:
1. All line breaks are converted inside JSON string:
```json
{ "maskMe": "multi\nline" }
```
->
```java
ValueMaskers.withTextFunction(value -> System.out.println(value.contains("\n")))    // false
ValueMaskers.withTextFunction(value -> System.out.println(value.contains("\\n")))   // true
```
2. Returning a line break would result in invalid JSON:
```java
ValueMaskers.withTextFunction(value -> "*\n*")
```
->
```json
{ "maskMe": "*
*" }
```
3. Resulting strings ending with an escape character or skipping an escape character, which would also produce invalid JSON
```json
{ "maskMe": "multi\nline" }
```
with
```java
ValueMaskers.withTextFunction(value -> value.substring(0, 6))
```
->
```json
{ "maskMe": "multi\" }
```

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
@@ -53,7 +53,9 @@ public interface ValueMaskerContext {
     int countNonVisibleCharacters(int fromIndex, int length);
 
     /**
-     * Returns a string representation of the original JSON value.
+     * Returns a string representation of the original JSON value. 
+     * <p>
+     * Note: this INCLUDES the opening and closing quotes for string values 
      */
     String asString(int fromIndex, int length);
 }
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
@@ -172,12 +172,11 @@ public final class ValueMaskers {
      * Masks a target value with the provided {@link Function}. The target value (as raw JSON literal) is passed into
      * the function as a string regardless of the JSON type (string, numeric or a boolean). In case the target value is
      * a JSON string the value the function will receive a JSON encoded value with the quotes as it appears in the JSON
-     * with line breaks converted to \n, special characters like " or \ escaped with a backslash (\).
+     * with line breaks encoded as  \n, special characters like " or \ escaped with a backslash (\).
      *
-     * <p>Consequently the return value of the provided function must be a valid JSON literal (of any JSON type)
-     * with respective JSON encoding, otherwise the masking will result in an invalid JSON.
+     * <p>Consequently, the return value of the provided function must be a valid JSON encoded literal (of any JSON type), otherwise the masking will result in an invalid JSON.
      *
-     * <p>The table below contains couple of examples for the masking
+     * <p>The table below contains a couple examples for the masking
      * <table>
      *   <caption>Examples of using withRawValueFunction</caption>
      *   <tr>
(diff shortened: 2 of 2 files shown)


src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
```diff
@@ -53,8 +53,7 @@ public interface ValueMaskerContext {
     int countNonVisibleCharacters(int fromIndex, int length);
 
     /**
-     * Returns a textual representation of the original value. For string values, the opening and
-     * closing quotes are EXCLUDED.
+     * Returns a string representation of the original JSON value.
```
Breus, line 56:
```suggestion
     * Returns a string representation of the original JSON value. 
     * <p>
     * Note: this INCLUDES the opening and closing quotes for string values 
```

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -169,25 +169,67 @@ public static ValueMasker.StringMasker email(int keepPrefixLength, int keepSuffi
...
 
     /**
-     * Masks a target value with the provided {@link Function}. The target value is passed into the
-     * function as a text, regardless of the type (string, numeric or a boolean), however any
-     * non-null returned value from the function will always be a JSON string (with quotes added
-     * automatically).
+     * Masks a target value with the provided {@link Function}. The target value (as raw JSON literal) is passed into
+     * the function as a string regardless of the JSON type (string, numeric or a boolean). In case the target value is
+     * a JSON string the value the function will receive a JSON encoded value with the quotes as it appears in the JSON
+     * with line breaks converted to \n, special characters like " or \ escaped with a backslash (\).
```
Breus, line 175:
```suggestion
     * with line breaks encoded as  \n, special characters like " or \ escaped with a backslash (\).
```

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -169,25 +169,67 @@ public static ValueMasker.StringMasker email(int keepPrefixLength, int keepSuffi
...
+     *
+     * <p>Consequently the return value of the provided function must be a valid JSON literal (of any JSON type)
+     * with respective JSON encoding, otherwise the masking will result in an invalid JSON.
+     *
+     * <p>The table below contains couple of examples for the masking
```
Breus, line 180:
```suggestion
     * <p>The table below contains a couple examples for the masking
```

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -169,25 +169,67 @@ public static ValueMasker.StringMasker email(int keepPrefixLength, int keepSuffi
...
+     *   <tr>
+     *     <td>{@code { "maskMe": "Andrii \"Juice\" Pilshchykov" }}
+     *     <td>{@code value -> value.replaceAll("\\\"", "(quote)")}
+     *     <td>{@code { "maskMe": "Andrii (quote)Juice(quote) Pilshchykov" }} (valid JSON)
+     * </table>
      *
      * <p>Note: usually {@link ValueMasker} operates on a byte level without parsing JSON values
      * into intermediate objects, however this implementation will have to allocate a {@link String}
      * before passing it into the function and then turn it back into a byte array for replacement.
      */
```
Breus, line 224:
```suggestion
     * <p>Note: usually the {@link ValueMasker} operates on a byte level without parsing JSON values
     * into intermediate objects. This implementation, however,  needs to allocate a {@link String}
     * before passing it into the function and then turn it back into a byte array for the replacement, which introduces 
     * some performance overhead.
     */
```

Breus (approved):
Left a couple suggestions for the documentation. For the rest, LGTM

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
gavlyukovskiy: Those as well and there are quite some more + all unicode, so I didn't want to include them all

Breus: Then I would mention that that is just an example

gavlyukovskiy: Darn, it got automerged because committing a suggestion resolves the thread 🙃

Breus: 😂😂

PR #104, opened 2024-03-27
Add actual withTextFunction that decodes / encodes JSON values

gavlyukovskiy:
This PR adds a function that will actually decode / encode the JSON string into Java string and back.
Decoding would be simple if not for `\uXXXX` that need to be decoded. It also doesn't help that we're working on byte level, so that I needed to convert 4 byte -> int -> 1-2 chars -> 1-4 bytes. That code is like magic at this point because 
1. ~~ChatGPT wrote most of it (from many tries)~~ took me a while to actually understand the original code, so instead now it's almost a full copy of `String#encodeUTF8_UTF16` 😝 
2. I wrote the some of it after finding bunch of problems

In my defense, it does match (somewhat) the magic done in [Jackson](https://github.com/FasterXML/jackson-core/blob/2.18/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java#L2123-L2134) and [String#encodeUTF8_UTF16](https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/java.base/share/classes/java/lang/String.java#L1322-L1374) 😅 

Encoding is simple as it's only escapes necessary characters without encoding unicode symbols into `\uXXXX` (that's allowed)

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/InvalidJsonException.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/InvalidJsonException.java
@@ -13,6 +13,7 @@ public class InvalidJsonException extends RuntimeException {
     public InvalidJsonException(String message, Throwable cause) {
         super(message, cause);
     }
+
     public InvalidJsonException(String message) {
         super(message);
     }
--- a/src/main/java/dev/blaauwendraad/masker/json/InvalidJsonException.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/InvalidJsonException.java
@@ -13,4 +13,7 @@ public class InvalidJsonException extends RuntimeException {
     public InvalidJsonException(String message, Throwable cause) {
         super(message, cause);
     }
+    public InvalidJsonException(String message) {
+        super(message);
+    }
 }
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
@@ -59,5 +59,13 @@ public interface ValueMaskerContext {
      */
     String asString(int fromIndex, int length);
 
+    /**
+     * Create an {@link InvalidJsonException} with the given message and index relative to the value (i.e. an index
+     * between {@code 0} and {@link ValueMaskerContext#byteLength()}).
+     *
+     * @param message error message
+     * @param index relative index where the JSON contains invalid sequence
+     * @return the exception to be thrown
+     */
     InvalidJsonException invalidJson(String message, int index);
 }
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
@@ -58,4 +58,6 @@ public interface ValueMaskerContext {
      * Note: this INCLUDES the opening and closing quotes for string values 
      */
     String asString(int fromIndex, int length);
+
+    InvalidJsonException invalidJson(String message, int index);
 }
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
@@ -337,7 +337,7 @@ public final class ValueMaskers {
                 "withTextFunction (%s)".formatted(masker),
                 context -> {
                     String decodedValue;
-                    if (context.byteLength() > 0 && context.getByte(0) == '"') {
+                    if (context.getByte(0) == '"') {
                         int encodedIndex = 1; // skip opening quote
                         int valueEndIndex = context.byteLength() - 1; // skip closing quote
                         int decodedIndex = 0;
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
@@ -349,12 +349,12 @@ public final class ValueMaskers {
                                 originalByte = context.getByte(encodedIndex++);
                                 switch (originalByte) {
                                     // First, ones that are mapped
-                                    case 'b' -> decodedBytes[decodedIndex] = '\b';
-                                    case 't' -> decodedBytes[decodedIndex] = '\t';
-                                    case 'n' -> decodedBytes[decodedIndex] = '\n';
-                                    case 'f' -> decodedBytes[decodedIndex] = '\f';
-                                    case 'r' -> decodedBytes[decodedIndex] = '\r';
-                                    case '"', '/', '\\' -> decodedBytes[decodedIndex] = originalByte;
+                                    case 'b' -> decodedBytes[decodedIndex++] = '\b';
+                                    case 't' -> decodedBytes[decodedIndex++] = '\t';
+                                    case 'n' -> decodedBytes[decodedIndex++] = '\n';
+                                    case 'f' -> decodedBytes[decodedIndex++] = '\f';
+                                    case 'r' -> decodedBytes[decodedIndex++] = '\r';
+                                    case '"', '/', '\\' -> decodedBytes[decodedIndex++] = originalByte;
                                     case 'u' -> {
                                         // Decode Unicode character
                                         // the copy of String#encodeUTF8_UTF16 with the only difference that it
@@ -406,14 +406,12 @@ public final class ValueMaskers {
                                             decodedBytes[decodedIndex++] = (byte) (0x80 | ((c >> 6) & 0x3f));
                                             decodedBytes[decodedIndex++] = (byte) (0x80 | (c & 0x3f));
                                         }
-                                        continue;
                                     }
                                     default -> throw context.invalidJson("Unexpected character after '\\': '%s'".formatted((char) originalByte), encodedIndex);
                                 }
                             } else {
-                                decodedBytes[decodedIndex] = originalByte;
+                                decodedBytes[decodedIndex++] = originalByte;
                             }
-                            decodedIndex++;
                         }
                         decodedValue = new String(decodedBytes, 0, decodedIndex, StandardCharsets.UTF_8);
                     } else {
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
@@ -397,13 +397,13 @@ public final class ValueMaskers {
                                             } else {
                                                 decodedBytes[decodedIndex++] = (byte) (0xf0 | (uc >> 18));
                                                 decodedBytes[decodedIndex++] = (byte) (0x80 | ((uc >> 12) & 0x3f));
-                                                decodedBytes[decodedIndex++] = (byte) (0x80 | ((uc >>  6) & 0x3f));
+                                                decodedBytes[decodedIndex++] = (byte) (0x80 | ((uc >> 6) & 0x3f));
                                                 decodedBytes[decodedIndex++] = (byte) (0x80 | (uc & 0x3f));
                                             }
                                         } else {
                                             // 3 bytes, 16 bits
                                             decodedBytes[decodedIndex++] = (byte) (0xe0 | (c >> 12));
-                                            decodedBytes[decodedIndex++] = (byte) (0x80 | ((c >>  6) & 0x3f));
+                                            decodedBytes[decodedIndex++] = (byte) (0x80 | ((c >> 6) & 0x3f));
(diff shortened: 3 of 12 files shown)


src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,172 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+                                                }
+                                            }
+                                            if (uc < 0) {
+                                                // default String behaviour is to replace invalid surrogate pairs
+                                                // with the character '?', but from the JSON perspective,
+                                                // it's better to throw an exception
+                                                throw context.invalidJson("Invalid surrogate pair '%s', expected '\\uXXXX\\uXXXX'"
+                                                        .formatted(context.asString(valueStartIndex, encodedIndex - valueStartIndex)), valueStartIndex);
+                                            } else {
+                                                decodedBytes[decodedIndex++] = (byte) (0xf0 | ((uc >> 18)));
```
gavlyukovskiy, line 398:
```suggestion
                                                decodedBytes[decodedIndex++] = (byte) (0xf0 | (uc >> 18));
```

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,172 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+                                                decodedBytes[decodedIndex++] = (byte) (0x80 | ((uc >> 12) & 0x3f));
+                                                decodedBytes[decodedIndex++] = (byte) (0x80 | ((uc >>  6) & 0x3f));
+                                                decodedBytes[decodedIndex++] = (byte) (0x80 | (uc & 0x3f));
+                                            }
+                                        } else {
+                                            // 3 bytes, 16 bits
+                                            decodedBytes[decodedIndex++] = (byte) (0xe0 | ((c >> 12)));
```
gavlyukovskiy, line 405:
```suggestion
                                            decodedBytes[decodedIndex++] = (byte) (0xe0 | (c >> 12));
```

gavlyukovskiy (commented):
Code suggestions

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,172 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+     * into intermediate objects. This implementation, however,  needs to allocate a {@link String}
+     * before passing it into the function and then turn it back into a byte array for the replacement,
+     * which introduces some performance overhead.
+     */
+    public static ValueMasker.AnyValueMasker withTextFunction(Function<String, String> masker) {
+        return describe(
+                "withTextFunction (%s)".formatted(masker),
+                context -> {
+                    String decodedValue;
+                    if (context.byteLength() > 0 && context.getByte(0) == '"') {
```
gavlyukovskiy, line 340:
```suggestion
                    if (context.getByte(0) == '"') {
```
context.byteLength can never be 0

src/main/java/dev/blaauwendraad/masker/json/InvalidJsonException.java
```diff
@@ -13,4 +13,7 @@ public class InvalidJsonException extends RuntimeException {
     public InvalidJsonException(String message, Throwable cause) {
         super(message, cause);
     }
+    public InvalidJsonException(String message) {
```
gavlyukovskiy, line 16:
```suggestion
    }

    public InvalidJsonException(String message) {
```

donavdey (approved):

src/main/java/dev/blaauwendraad/masker/json/util/Utf8Util.java
```diff
@@ -31,6 +31,39 @@ public static int getCodePointByteLength(byte input) {
...
+        if (hexByte >= 48 && hexByte <= 57) {
+            return hexByte; // a digit from 0 to 9
+        }
+        if (hexByte >= 65 && hexByte <= 70) {
+            return hexByte; // a character from A to F
+        }
+        if (hexByte >= 97 && hexByte <= 102) {
+            return hexByte; // a character from a to f
+        }
+        throw new IllegalArgumentException("Invalid hex character '%s'".formatted((char) hexByte));
```
Breus, line 63: Shouldn't this result in an (subclass of) InvalidJsonException (such as `InvalidJsonCharacterEncoding`) according to our contract?

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,182 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+                    byte[] replacementBytes = maskedValue.getBytes(StandardCharsets.UTF_8);
+                    context.replaceBytes(0, context.byteLength(), replacementBytes, 1);
+                });
+    }
+
+
+    /**
+     * Masks a target value with the provided {@link Function}. The target value (as textual representation of a
+     * JSON value) is passed into the function as a string regardless of the JSON type (string, numeric or a boolean).
+     * In case original value was a JSON string, the function will receive a decoded string value without the quotes.
```
Breus, line 290:
```suggestion
     * In case the original value was a JSON string, the function will receive a decoded string value without the quotes.
```

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,182 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
-                    context.replaceBytes(0, context.byteLength(), maskedValue.getBytes(StandardCharsets.UTF_8), 1);
+                    byte[] replacementBytes = maskedValue.getBytes(StandardCharsets.UTF_8);
+                    context.replaceBytes(0, context.byteLength(), replacementBytes, 1);
+                });
+    }
+
+
+    /**
+     * Masks a target value with the provided {@link Function}. The target value (as textual representation of a
+     * JSON value) is passed into the function as a string regardless of the JSON type (string, numeric or a boolean).
```
Breus, line 289:
```suggestion
     * JSON value) is passed into the function as a string, regardless of the JSON type (string, numeric or a boolean).
```

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,182 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+     * In case original value was a JSON string, the function will receive a decoded string value without the quotes.
+     *
+     * <p>A non-null return value of the provided function will be encoded into a JSON string regardless of the
+     * JSON type of the original value. Any character that MUST be escaped (as per RFC 8259, section 7) will be escaped,
+     * characters that MAY be escaped (as per RFC 8259) WILL NOT be escaped.
```
Breus, line 294:
```suggestion
     * JSON type of the original value. Any character that MUST be escaped (as per RFC 8259, section 7) will be escaped. 
     * Characters that MAY be escaped (as per RFC 8259) WILL NOT be escaped.
```

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,182 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+     * into intermediate objects. This implementation, however,  needs to allocate a {@link String}
+     * before passing it into the function and then turn it back into a byte array for the replacement,
+     * which introduces some performance overhead.
+     */
+    public static ValueMasker.AnyValueMasker withTextFunction(Function<String, String> masker) {
+        return describe(
+                "withTextFunction (%s)".formatted(masker),
+                context -> {
+                    String decodedValue;
+                    if (context.getByte(0) == '"') {
```
Breus, line 340: Can this if/else be flipped for better readability? Now by the time you see `else` you might have forgotten that it was (not string) basically

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,182 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+                        int encodedIndex = 1; // skip opening quote
+                        int valueEndIndex = context.byteLength() - 1; // skip closing quote
+                        int decodedIndex = 0;
```
Breus, line 343: At this point of reading, it is completely unclear what `encodedIndex` and `decodedIndex` are and why encodedIndex is one higher. Would it be possible to have more descriptive names? 

I will see if I can think of a way to make this code more clear

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,182 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+                        // 1. for every encoded character (2 bytes), the output is the character without escape - 1 byte
+                        // 2. for every unicode encoded character (6 bytes), the output character is within 1-3 bytes
+                        // 3. for the pair of unicode encoded surrogates (12 bytes), the output is always 4 bytes
+                        // which means that if any escape is present, the decodedBytes will have some null characters
+                        // at the tail, which are cut from the resulting string
+                        byte[] decodedBytes = new byte[context.byteLength()];
+                        while (encodedIndex < valueEndIndex) {
+                            byte originalByte = context.getByte(encodedIndex++);
+                            // next character is escaped, removing the backslash
+                            if (originalByte == '\\') {
```
Breus, line 355: Also would be nice to switch this if/else to not lose track of context for the else. Alternatively, moving the entire block of code inside the else to a new private method might be nice for readability.

src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
```diff
@@ -58,4 +58,14 @@ public interface ValueMaskerContext {
      * Note: this INCLUDES the opening and closing quotes for string values 
      */
     String asString(int fromIndex, int length);
+
+    /**
+     * Create an {@link InvalidJsonException} with the given message and index relative to the value (i.e. an index
+     * between {@code 0} and {@link ValueMaskerContext#byteLength()}).
```
Breus, line 64: Does this really make sense to give an index relative to the value in the message? Is it clear from a stacktrace which part is invalid exactly?

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -258,7 +278,186 @@ public static ValueMasker.AnyValueMasker withRawValueFunction(Function<String, S
...
+                            char character = maskedValue.charAt(i);
+                            // escape all characters that need to be escaped, unicode character do not have to be
+                            // transformed into \ u form
+                            switch (character) {
+                                case '\b' -> encoded.append("\\b");
+                                case '\t' -> encoded.append("\\t");
+                                case '\n' -> encoded.append("\\n");
+                                case '\f' -> encoded.append("\\f");
+                                case '\r' -> encoded.append("\\r");
+                                case '"', '\\' -> encoded.append("\\").append(character);
```
Breus, line 452: What about `\`? Is that one of those characters that don't have to be encoded?

gavlyukovskiy: `\` and `"` always must be encoded, `/` might be encoded, but we don't do that

src/main/java/dev/blaauwendraad/masker/json/util/Utf8Util.java
gavlyukovskiy: Eh? I guess it could be, but I don't think we have to go there with differentiating why exactly json is invalid. It's not like it would matter for the user in terms of handling the error

Breus: Sure, so just regular Invalid Json Exception

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
Breus: Cool

src/main/java/dev/blaauwendraad/masker/json/ValueMaskerContext.java
gavlyukovskiy: It won't be that much clear, for example if a string value has an invalid sequence, then the stack trace will be at `maskString` and nothing more, so it won't be possible to know which string is that even

Breus (approved):

PR #108, opened 2024-03-29
Add test showing repeating each digit of a number with an asterisk

Breus:


src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -112,11 +112,38 @@ public static ValueMasker.NumberMasker eachDigitWith(int digit) {
...
+     * character).
+     *
+     * <p> For example, {@literal "maskMe": 12345 -> "maskMe": "*****"}.
+     * <p> Or, for example {@literal "maskMe": 123 -> "maskMe": "NoNoNo"}.
+     */
+    public static ValueMasker.NumberMasker eachDigitWith(String value) {
+        return describe(
+                "every digit as string: %s".formatted(value),
+                context -> {
+                    byte[] maskValueBytes = value.getBytes(StandardCharsets.UTF_8);
```
gavlyukovskiy, line 131: Do this outside of lambda so it is caches for all masked values

gavlyukovskiy (approved):

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
Breus: Nice catch! I moved both lines to outside of the lambda

PR #116, opened 2024-04-03
Optimize json path lookup by avoiding traversing the trie from the ro…

donavdey:
…ot every time.


src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -46,13 +45,12 @@ public byte[] mask(byte[] input) {
...
-                        -1,
-                        Collections.emptyIterator()
-                );
+                // A special case of "$" JSONPath key.
+                maskingState.expandCurrentJsonPath(keyMatcher.getJsonPathRootNode());
+                if (maskingState.getCurrentJsonPathNode().endOfWord && maskingConfig.isInAllowMode()) {
+                    return input;
+                }
+                keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), -1, -1, maskingState.getCurrentJsonPathNode());
             }
```
gavlyukovskiy, line 54: With this amount of code handling this particular edge case (`allowJsonPaths("$")`), that also doesn't make practical sense, I wonder whether we should actually disallow that? Basically it means "mask nothing" and we already disallow similar situation with `maskKeys(Set.of())` throwing an exception

donavdey: Makes sence. Disallowed $ in allow mode.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -21,18 +19,19 @@ final class MaskingState implements ValueMaskerContext {
...
 
     /**
-     * Current JSONPath is represented by a dequeue of segment references.
+     * Current JSONPath is represented by a stack of segment references.
+     * A stack is implemented with an array of the trie nodes that reference the end of the segment
      */
-    private final Deque<JsonPathNode> currentJsonPath;
+    private KeyMatcher.TrieNode[] currentJsonPath = null;
+    private int currentJsonPathIndex = -1;
+    private int currentJsonPathCapacity = 100;
```
gavlyukovskiy, line 27: This can be `currentJsonPath.length`, no need to track it separately

donavdey: Nice one, fixed.

PR #119, opened 2024-04-05
Allow overriding masking config for more specific keys for jsonpath

donavdey:


src/test/java/dev/blaauwendraad/masker/json/path/JsonPathParserTest.java
```diff
@@ -100,10 +100,7 @@ private static Stream<Set<String>> ambiguousJsonPaths() {
                 Set.of("$.a.b.c.f", "$.*.*.*.u"),
                 Set.of("$.*.b.c", "$.q.w.e", "$.*.d.f"),
                 Set.of("$.a.b.c", "$.d.*.*.f", "$.d.*.v.f"),
-                Set.of("$.a.b.*.*.d", "$.a.b.*.c"),
```
gavlyukovskiy, line 103: This seems to be the one we should still not allow, no?

src/test/resources/test-json-path.json
```diff
@@ -1458,5 +1458,163 @@
       },
       "otherKey": "***"
     }
+  },
+  {
+    "maskingConfig": {
+      "maskJsonPaths": [
+        "$.json.path"
+      ],
+      "maskJsonPathsAlternatively": [
```
gavlyukovskiy, line 1467: Maybe instead of using `maskJsonPathsAlternatively`, you could make it that `maskJsonPaths` could also accept a json object of format `"maskJsonPaths": ["a", { "jsonPaths": ["b", "c"], "maskNumbersWith": ...}`?

If that's to difficult, then I'd rather look into a specific test that tests such behavior for regular keys (don't remember the name, maybe CustomJsonMaskingTest or something like that)

src/test/java/dev/blaauwendraad/masker/json/path/JsonPathParserTest.java
gavlyukovskiy: Basically when you have the path `a.b.x.c` then it's still ambiguous

donavdey: That's "diff" issue, this pair is 4 lines below

src/test/resources/test-json-path.json
donavdey: Should not be difficult, I will do that

donavdey: Implemented.

src/test/java/dev/blaauwendraad/masker/json/JsonMaskerTestUtil.java
```diff
@@ -49,7 +43,13 @@ private static void applyConfig(JsonNode jsonMaskingConfig, JsonMaskingConfig.Bu
             JsonNode value = e.getValue();
             switch (key) {
                 case "maskKeys" -> builder.maskKeys(asSet(value, JsonNode::asText));
-                case "maskJsonPaths" -> builder.maskJsonPaths(asSet(value, JsonNode::asText));
+                case "maskJsonPaths" -> StreamSupport.stream(value.spliterator(), false).forEach(node -> {
```
gavlyukovskiy, line 46: Nice, can you do the same for maskKeys? Should be the same logic

gavlyukovskiy (approved):
LGTM, added a small comment

src/test/java/dev/blaauwendraad/masker/json/JsonMaskerTestUtil.java
donavdey: Done

PR #122, opened 2024-04-12
Make sure pipelines run from forks

gavlyukovskiy:
To make sure that workflow run and finish correctly, we're using the [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target). This runs the PR with full privileges, but it doesn't allow modifying the pipeline itself from the PR.
It's safe to run the benchmark pipeline in `pull_request_target` as it doesn't have access to the secrets, however our `build` pipeline explicitly propagates secrets when runs gradle, therefore it would be possible for the PR pipeline to do `System.getenv` and expose the secret values.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
       - master
     tags-ignore:
       - '*'
-  pull_request:
+  pull_request_target:
     branches:
       - master
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,9 +5,10 @@ on:
       - master
     tags-ignore:
       - '*'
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
       - master
```
Breus, line 10: What exactly does workflow_dispatch do?

gavlyukovskiy: That's a manual trigger, so you can go to actions tab and trigger a workflow on any branch. I'm not sure it will work for forks, but we'll see (and it doesn't hurt 😅)

PR #123, opened 2024-04-12
Migrate to JSpecify and NullAway

gavlyukovskiy:


src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -292,10 +291,9 @@ private static void stepOverWhitespaceCharacters(MaskingState maskingState) {
      */
     private static void stepOverNumericValue(MaskingState maskingState) {
         // step over the first numeric character
```
Breus, line 293:
```suggestion
```

Breus: That comment doesn't make sense after your change

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -249,11 +250,20 @@ static class TrieNode {
...
         boolean negativeMatch = false;
+
+        @Nullable
+        public TrieNode child(byte b) {
+            return children[b + BYTE_OFFSET];
+        }
+
+        public void add(byte b, TrieNode child) {
+            children[b + BYTE_OFFSET] = child;
+        }
```
Breus, line 267: I don't really like that these methods don't have JavaDoc. The whole KeyMatcher is already complex enough including good docs (at least for me) 😅

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -426,7 +427,7 @@ the byte encoded unicode character (\\u + 4 hex bytes) to prevent possible Array
                                                 decodedBytes[decodedIndex++] = (byte) (0x80 | (unicodeHexBytesAsChar & 0x3f));
                                             }
                                         } catch (IllegalArgumentException e) {
-                                            throw context.invalidJson(e.getMessage(), valueStartIndex);
+                                            throw context.invalidJson(Objects.requireNonNull(e.getMessage()), valueStartIndex);
```
Breus, line 430: How can `e.getMessage()` ever be `null` ?

src/main/java/dev/blaauwendraad/masker/json/config/KeyMaskingConfig.java
```diff
@@ -61,6 +62,7 @@ public String toString() {
                 .formatted(maskStringsWith, maskNumbersWith, maskBooleansWith);
     }
 
+    @NullUnmarked
```
Breus, line 65: Why exactly do we mix NullUnmarked and NullMarked (on packages)? Shouldn't we have `Nullunmarked` everywhere? I think I'm missing something

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
Breus: Especially since they are public (do they need to be?)

Breus:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=Breus_json-masker&pullRequest=123) **Quality Gate passed**
> 
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=Breus_json-masker&pullRequest=123&resolved=false&inNewCodePeriod=true)  [0 Accepted issues](https://sonarcloud.io/component_measures?id=Breus_json-masker&pullRequest=123&metric=new_accepted_issues&view=list)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=Breus_json-masker&pullRequest=123&resolved=false&inNewCodePeriod=true)  [96.9% Coverage on New Code](https://sonarcloud.io/component_measures?id=Breus_json-masker&pullRequest=123&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=Breus_json-masker&pullRequest=123&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=Breus_json-masker&pullRequest=123)

96.9% coverage?! Absolutely unacceptable @gavlyukovskiy 😅

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: The class itself is package private, so I could make them that as well, I didn't put too much thought into it just got tired of messing with the offsets 😅

Breus: Yea, let's make them package private and add some docs

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
gavlyukovskiy: Technically, the exception message can be null, which is why it complains. But in this particular case, we're the ones throwing IAE, so it's safe to assert it

Breus: Ugh, bit annoying imo to change code just to get the stupid analyser happy, but sure

src/main/java/dev/blaauwendraad/masker/json/config/KeyMaskingConfig.java
Breus: We can make the builder methods non-nullable, right?

gavlyukovskiy: Okay, it actually did revert all the methods, so I actually marked the fields nullable instead 👍 Nice catch

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
gavlyukovskiy: 🤷

gavlyukovskiy: At least that's an exceptional situation :)

src/main/java/dev/blaauwendraad/masker/json/JsonMasker.java
```diff
@@ -17,7 +16,6 @@ public interface JsonMasker {
      * @param targetKeys the key(s) to target
      * @return the {@link JsonMasker} instance
      */
-    @Nonnull
```
Breus, line 20: @gavlyukovskiy I just thought a bit more about it. I this we should now, at least to the main methods that form the API of the library, but in the `@return` JavaDoc that it will not return null.

Before developers could see this from the ` @Nonnull` annotations. Now they need to know that we use NullAway (which most users probably won't actually look at)

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -172,7 +172,7 @@ void backtrackCurrentJsonPath() {
     /**
      * Returns the TrieNode that references the end of the latest segment in the current jsonpath
      */
-    public KeyMatcher.TrieNode getCurrentJsonPathNode() {
+    public KeyMatcher.@Nullable TrieNode getCurrentJsonPathNode() {
```
Breus, line 175: Woah, this looks so odd. Didn't even know this was possible in Java 😂

src/main/java/dev/blaauwendraad/masker/json/JsonMasker.java
gavlyukovskiy: I honestly think we would be overdoing it and making the docs harder to read. Even if the developer doesn't use an IDE, that would scream about redundant null-check, it should be obvious that a factory method `getMasker` doesn't return `null`, adding that to the doc is quite useless in my opinion.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
gavlyukovskiy: Yeah, I didn't quite get why it needs to be done like that, I tried to do it on the whole field and it said something along the lines of "outer type can never be null for static inner class". While it kinda makes sense to apply it to inner type, I also don't get how that outer type can ever be null. I always thought that these outer types are kind of a namespace for static classes, you can import such class directly if you want, but can leave it be for more clarity (I.e. `Entry` vs `Map.Entry`), I always preferred the second, but it's so ugly with these annotations there 😅

Breus (approved):

PR #133, opened 2024-04-15
Make sure that all masks are JSON encoded

gavlyukovskiy:


src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -87,9 +87,11 @@ public static ValueMasker.AnyValueMasker withNull() {
      * \}uDCA9'), will only count as a single character in the masked value.
      */
     public static ValueMasker.StringMasker eachCharacterWith(String value) {
-        byte[] replacementBytes = value.getBytes(StandardCharsets.UTF_8);
+        String encoded = Utf8Util.jsonEncode(value);
+        String replacement = encoded.substring(1, encoded.length() - 1); // remove the quotes
+        byte[] replacementBytes = replacement.getBytes(StandardCharsets.UTF_8);
```
Breus, line 92: Can we not just have a method that doesn't add quotes around a string and still encodes and the one that does calls this + adds quotes? 

I find this quite ugly to first add them just because the existing method does so and then substringing to remove them 😅 It is also less obvious to read.

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -130,17 +132,18 @@ public static ValueMasker.NumberMasker eachDigitWith(int digit) {
      * <p> Or, for example {@literal "maskMe": 123 -> "maskMe": "NoNoNo"}.
      */
     public static ValueMasker.NumberMasker eachDigitWith(String value) {
-        byte[] maskValueBytes = value.getBytes(StandardCharsets.UTF_8);
-        int maskValueBytesLength = maskValueBytes.length;
+        String encoded = Utf8Util.jsonEncode(value);
+        String replacement = encoded.substring(1, encoded.length() - 1); // remove the quotes
```
Breus, line 136: Same here as I mentioned before

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
gavlyukovskiy: I hoped you wouldn't notice 😅 I guess I can add a boolean parameter there

Breus: 😂😂

Breus: I have a slight preference for two methods were the one with enclosing quotes calls the one without and adds quotes. It's better readible outside an IDE imo

Breus: "jsonEncode()" and "jsonEncodeWithEnclosingQuotes" or something along those lines

gavlyukovskiy: That method is quite large to just copy it with and without quotes. I was thinking about never adding the quotes in the util method and instead to do that inside withTextFunction, but that one is called for every value, while these two are only at configuration time.

Breus: Ahhh, I see. Then the boolean or keeping it like this is fine

Breus: @gavlyukovskiy lemme know if you want to keep it like this or introduce the boolean

Breus (approved):

PR #140, opened 2024-05-04
Add tests from JSONTestSuite

### LINKED ISSUE #121: Integrate JSONTestSuite
As suggested on reddit, we should make sure that json-masker can parse (and possibly mask) all inputs from https://github.com/nst/JSONTestSuite

gavlyukovskiy:
This PR adds all tests from `JSONTestSuite`. That suite has 3 types of tests:
- `y_*` are valid JSONs
- `i_*` are invalid JSONs that are allowed to be parsed, usually some large numbers and/or invalid UTF-8 sequences
- `n_*` are invalid JSONs, for this the normal parser should fail, however we can process all of them except 2 that have 100k nesting level (P.S. Don't try opening them in IDEA 😄)

I've copied all of them to our repository and established following test cases
#### mustPassSuiteWithNoopMaskerShouldBeEquivalentToJackson
`y_*` tests must pass. I'm using `ValueMaskers.withRawValueFunction(value -> value)` as a more complicated noop masker to make sure that we always correctly track values, transform them to strings and back to bytes. The results of this are equivalent when parsed by Jackson

#### mayPassSuiteWithNoopMaskerShouldNotFail
`i_*` tests must not fail with and ideally should be equivalent to how Jackson parses it (if it can). There are 4 exceptions when `java.lang.String` treats invalid UTF-8 characters differently than Jackson does

#### mustFailSuiteWithNoopMaskerShouldOnlyFailWithInvalidJsonException
`n_*` tests must fail for a normal JSON parser, however we allow it not to fail, but if it does fail, then it must be with `InvalidJsonException`.  With `ValueMaskers.withRawValueFunction` (or default masking options) there are only 2 test cases that are failing due to internal `StackOverFlowError`.

#### shouldMaskAllTestCasesPredictably

For all test files (including `n_*`) I've simply asserted the masker json we currently produce.

#### mustPassSuiteWithNoopTextFunction

Verifies that all test files that contain correct UTF-8 sequences we properly decode it in `withTextValueFunction`. There are 26 `i_*` test files where we fail on converting invalid sequences.

For the `withTextValueFunction` function the question is whether we should continue to fail on those 26 test cases or take the behavior of `java.lang.String` of converting most invalid sequences into `�` and passing that to the user provided function?

### Fixes

This PR has couple of fixes for cases when we get out of bounds or throw internal exception instead of `InvalidJsonException`. All of the fixes only apply to `n_*` and `i_*` test cases, nothing critical was found during the testing 😃 

### Hard decisions time!

The main problem is what to do with `n_*` test cases? 

The main problem is that we're not actually parsing the values so we do not reject things like invalid numbers or invalid escapes (except for `withTextValueFunction`. We also do not fail on invalid json structures - for example in arrays we don't strictly need to see a comma and ignore anything unknown and mask values if we detect them. With that for most of the test cases we actually mask them in more or less predictable ways, for example 
- `[""` -> `["***"` (`n_array_unclosed.json`)
- `[   , ""]` -> `[   , "***"]` (`n_array_missing_value.json`)

However some test cases are leaking information or giving weird results
- `abc` -> `abc` (`n_string_single_string_no_double_quotes.json`) when string doesn't have quotes then we simply leave it as is 
- `['single quote']` -> `['single quo"&&&"` (`n_string_single_quote.json`) we ignore the value started with a single quote and accidentally think that we've found a boolean to mask

In `shouldMaskAllTestCasesPredictably` I've asserted the current status, so it would be useful to go over them and decide whether we should do anything special about each one of those.

Closes #121


gavlyukovskiy:
@Breus looks like that badge is going to get a little bit more impressive 😄 
image

donavdey (approved):
> The main problem is what to do with n_* test cases?

I'd suggest to clearly claim in the docs that the masker is not a RFC 8259 compliant JSON parser, so the output for non-valid JSONs may be unpredictable.

Breus (approved):
LGTM. Let's address the invalid input handling (? vs throwing exception) in a different PR. Not too big of a deal

PR #141, opened 2024-05-25
Handle masking of unicode keys

gavlyukovskiy:
This PR adds handling for unicode encoded keys in the input:
```json
{
    "\u006D\u0061\u0073\u006B\u004D\u0065": "secret"
}
```
(encoded key `maskMe`)

The logic is pretty similar to one in `ValueMaskers.withTextFunction`, but unfortunately can't be unified due to differences in handling.

> [!NOTE]
> I did not apply the same logic to input keys when passed to `maskKeys` / `allowKeys` as there would be absolutely no reason to pass us keys in such a format


src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -37,8 +38,6 @@ final class KeyMatcher {
     private static final int SKIP_KEY_LOOKUP = -1;
     private final JsonMaskingConfig maskingConfig;
     private final TrieNode root;
-    // used for an optimization to remember all key length and return early if the key length is not known
-    private final boolean[] knownKeyLengthsInBytes = new boolean[256];
```
gavlyukovskiy, line 41: Can't use this (micro-)optimization anymore as every input character of 1-4 bytes can be matching 1-12 bytes of json key if that happens to be encoded.

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
```diff
@@ -394,7 +394,7 @@ public static ValueMasker.AnyValueMasker withTextFunction(Function<String, @Null
                                                 // surrogates which together form one unicode character.
                                                 int codePoint = -1;
                                                 if (Character.isHighSurrogate(unicodeHexBytesAsChar) // first surrogate must be the high surrogate
-                                                        && encodedIndex < context.byteLength() - 6 /* -6 for all bytes of
+                                                        && encodedIndex <= context.byteLength() - 6 /* -6 for all bytes of
```
gavlyukovskiy, line 397: Found this when copying the code into `KeyMatcher` and the latest encoded character wasn't matching properly due to incorrect condition. In this particular place it doesn't change anything because the latest character is always `"` so that condition is always true for the valid strings

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -178,14 +176,73 @@ KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, int keyOffset, int keyLeng
...
+                        && bytes[i + 1] == 'u'
+                    ) {
+                        char lowSurrogate = Utf8Util.unicodeHexToChar(bytes, i + 2);
+                        if (Character.isLowSurrogate(lowSurrogate)) {
+                            codePoint = Character.toCodePoint(unicodeHexBytesAsChar, lowSurrogate);
+                        }
+                    }
+                    if (codePoint < 0) {
+                        // the key contains invalid surrogate pair and won't be matched
+                        return null;
```
gavlyukovskiy, line 213: @Breus coming back to our discussion 😄 
Noticed that JSONTestSuite `i_object_key_lone_2nd_surrogate.json` started to fail as the key contains incomplete surrogate. Unlike in `withTextFunction` I think here it makes more sense to treat it as "key didn't match" (would still be masked in allow-mode).
For reference jackson parses that key as `�`

src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
gavlyukovskiy: Perhaps it's one more reason to adhere to jackson behavior in `withTextFunction` as well

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
gavlyukovskiy: Perhaps it's one more reason to adhere to jackson behavior in `withTextFunction` as well

donavdey (approved):
So I understood the general idea and what is going on in KeyMatcher, but got a bit lost in implementation details with byte shifts and byte masking. Nevertheless, I see the (very nice) tests completely cover possible cases, so I approve the PR.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -178,14 +176,73 @@ KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, int keyOffset, int keyLeng
...
-            node = node.child(b);
+            if (b == '\\' && bytes[i + 1] == 'u' && i <= offset + length - 6) {
+                char unicodeHexBytesAsChar = Utf8Util.unicodeHexToChar(bytes, i + 2);
+                i += 6;
+                if (unicodeHexBytesAsChar < 0x80) {
+                    // < 128 (in decimal) fits in 7 bits which is 1 byte of data in UTF-8
+                    node = node.child((byte) unicodeHexBytesAsChar);
+                } else if (unicodeHexBytesAsChar < 0x800) { // 2048 in decimal,
+                    // < 2048 (in decimal) fits in 11 bits which is 2 bytes of data in UTF-8
+                    node = node.child((byte) (0xc0 | (unicodeHexBytesAsChar >> 6)));
```
Breus, line 191: Can you say something about 0xc0 and 0x80 ORing?

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -178,14 +176,73 @@ KeyMaskingConfig getMaskConfigIfMatched(byte[] bytes, int keyOffset, int keyLeng
...
+                    if (node == null) {
+                        return null;
+                    }
+                    node = node.child((byte) (0x80 | (unicodeHexBytesAsChar & 0x3f)));
```
Breus, line 195: Why are you doing this here?

Breus (approved):

PR #146, opened 2024-06-20
Added ability to match snake case keys

gavlyukovskiy:


gavlyukovskiy:
After discussing it for a while, let's not bloat the lib with this. The snake case keys can be explicitly supplied as input without implicit transformations on our side.

PR #152, opened 2024-07-07
Add ValueMaskers implementation for IBAN

bramdelver:
Adds an `iban` function to ValueMaskers. This function will keep the country code, the bank identification number and (optionally) a suffix.

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/ValueMaskers.java
@@ -202,6 +202,42 @@ public final class ValueMaskers {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Breus:
@bramdelver honestly I don't think this library should natively support this as this is masker is not generic enough. If we support IBAN someone could also ask for SSN, BSN, National Insurance Number, some other contract number, etc.

The problem with that is twofold: 
First of all, it would massively clutter the code if we start adding all of those context-specific maskers
Second, each user might want to have those domain-specific maskers behave differently so we will have multiple ones at some point.

The reason we added email is because that is so ridiculously generic at this point that we thought it would make sense to provide something (customizable) out-of-the-box for that one.

PR #153, opened 2024-07-13
Implementation of radix trie

gavlyukovskiy:
[Radix tree](https://en.wikipedia.org/wiki/Radix_tree)

This PR implements radix trie that has reduced retained memory footprint. I hoped to also see some improvements in benchmarks that affect processing, but I don't see anything conclusive :shrug: Perhaps that is because of branching in the matching methods that is now a bit more complicated.

- [x] JSONPath tests are not passing, needs further investigation
- [x] Left a couple of TODOs with further optimizations
- [x] Look a bit more into why we didn't see any improvements for masking

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 import net.ltgt.gradle.errorprone.errorprone
+import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
 import org.sonarqube.gradle.SonarTask
 
 plugins {
@@ -187,6 +188,9 @@ tasks {
                 "UnusedVariable",
                 "WildcardImport",
             )
+            if (DefaultNativePlatform.getCurrentOperatingSystem().isWindows) {
+                disable("MisleadingEscapedSpace") // good stuff
+            }
             disable(
                 "StringCaseLocaleUsage",
                 "MissingSummary",
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -174,12 +174,17 @@ tasks {
                 "RedundantOverride",
                 "RedundantThrows",
                 "RemoveUnusedImports",
+                "DefaultCharset",
                 "UnnecessarilyFullyQualified",
                 "UnnecessarilyUsedValue",
                 "UnnecessaryBoxedAssignment",
                 "UnnecessaryBoxedVariable",
                 "UnnecessaryFinal",
                 "UnusedException",
+                "UnusedLabel",
+                "UnusedMethod",
+                "UnusedNestedClass",
+                "UnusedVariable",
                 "WildcardImport",
             )
             disable(
--- a/src/main/java/dev/blaauwendraad/masker/json/JsonPathState.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/JsonPathState.java
@@ -1,0 +1,103 @@
+package dev.blaauwendraad.masker.json;
+
+import org.jspecify.annotations.Nullable;
+
+import java.util.ArrayDeque;
+
+class JsonPathState {
+    private static final KeyMatcher.StatefulRadixTrieNode NULL_NODE = new KeyMatcher.StatefulRadixTrieNode(new KeyMatcher.RadixTrieNode(new byte[0], new byte[0]), 0);
+
+    private final KeyMatcher keyMatcher;
+    private final ArrayDeque<KeyMatcher.StatefulRadixTrieNode> jsonPathSegments = new ArrayDeque<>();
+
+    JsonPathState(KeyMatcher keyMatcher) {
+        this.keyMatcher = keyMatcher;
+        var root = keyMatcher.getRootNode();
+        if (!root.performChildLookup((byte) '$')) {
+            throw new IllegalStateException("JSONPath root node is null");
+        }
+        this.jsonPathSegments.push(new KeyMatcher.StatefulRadixTrieNode(root));
+        root.reset();
+    }
+
+    /**
+     * Expands current JSONPath with an array segment.
+     */
+    void pushArraySegment() {
+        jsonPathSegments.push(getWildcardNodeOrNullNode());
+    }
+
+    /**
+     * Expands current JSONPath with a value segment.
+     */
+    void pushKeyValueSegment(byte[] bytes, int keyOffset, int keyLength) {
+        jsonPathSegments.push(getKeyValueNodeOrNullNode(bytes, keyOffset, keyLength));
+    }
+
+    /**
+     * Backtracks current JSONPath to the previous segment.
+     */
+    void backtrack() {
+        jsonPathSegments.pop();
+    }
+
+    /**
+     * Traverse the trie node when entering an array. In order to match the array it has to be a wildcard.
+     * <p>For example:
+     * For a JSON like this {@code { "holder": [ { "maskMe": "secret" } } } the matching JSONPath has to be
+     * {@code '$.holder.*.maskMe'}, so that entering the array requires a wildcard node.
+     */
+    private KeyMatcher.StatefulRadixTrieNode getWildcardNodeOrNullNode() {
+        var current = currentNode();
+        if (current == null) {
+            return NULL_NODE;
+        }
+        try {
+            if (!current.performChildLookup((byte) '.')) {
+                return NULL_NODE;
+            }
+            if (current.isJsonPathWildcard()) {
+                current.performChildLookup((byte) '*');
+                return new KeyMatcher.StatefulRadixTrieNode(current);
+            }
+            return NULL_NODE;
+        } finally {
+            current.reset();
+        }
+    }
+
+    /**
+     * Traverse the trie node when entering a key-value. The matching can be done for the matching key, or through a wildcard ('*') JSONPath.
+     */
+    private KeyMatcher.StatefulRadixTrieNode getKeyValueNodeOrNullNode(byte[] bytes, int keyOffset, int keyLength) {
+        var current = currentNode();
+        if (current == null) {
+            return NULL_NODE;
+        }
+        try {
+            if (!current.performChildLookup((byte) '.')) {
+                return NULL_NODE;
+            }
+            if (current.isJsonPathWildcard()) {
+                current.performChildLookup((byte) '*');
+                return new KeyMatcher.StatefulRadixTrieNode(current);
+            } else {
+                var child = keyMatcher.traverseFrom(current, bytes, keyOffset, keyLength);
+                if (child != null) {
+                    return new KeyMatcher.StatefulRadixTrieNode(child);
+                }
+            }
+            return NULL_NODE;
+        } finally {
+            current.reset();
+        }
+    }
+
+    KeyMatcher.@Nullable StatefulRadixTrieNode currentNode() {
+        var peek = jsonPathSegments.peek();
+        if (peek == NULL_NODE) {
+            return null;
+        }
+        return peek;
+    }
+}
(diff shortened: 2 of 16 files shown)


gavlyukovskiy:
Was testing on of the benchmarks locally, the key matching actually got faster by quite a bit:

#### Before:
```text
Benchmark                         (caseSensitive)  (mode)   Mode  Cnt         Score   Error  Units
KeyMatcherBenchmark.matchAllKeys            false    mask  thrpt    2  16196074.837          ops/s
KeyMatcherBenchmark.matchAllKeys            false   allow  thrpt    2  16032336.619          ops/s
KeyMatcherBenchmark.matchAllKeys             true    mask  thrpt    2  13108860.366          ops/s
KeyMatcherBenchmark.matchAllKeys             true   allow  thrpt    2  13075189.300          ops/s
```
#### After:
```text
Benchmark                         (caseSensitive)  (mode)   Mode  Cnt         Score   Error  Units
KeyMatcherBenchmark.matchAllKeys            false    mask  thrpt    2  20520848.607          ops/s
KeyMatcherBenchmark.matchAllKeys            false   allow  thrpt    2  20466876.402          ops/s
KeyMatcherBenchmark.matchAllKeys             true    mask  thrpt    2  20085632.774          ops/s
KeyMatcherBenchmark.matchAllKeys             true   allow  thrpt    2  19213484.306          ops/s
```

so it looks like the reason we don't see much improvement on the existing benchmarks is because we have `maskedKeyProbability = 0.1` and we just don't have a lot of things to mask. Will include this benchmark in the PR.

donavdey (approved):
I guess there is a room to optimize JsonMakser instance creation time if we set out minds to do so, but otherwise the implementation seems pretty nice.
As discussed before, I will do something about the JSONPath stack so that we don't have to have a separate StatefulRadixTrieNode implementation

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -46,6 +46,7 @@ final class KeyMatcher {
     private static final int SKIP_KEY_LOOKUP = -1;
     private final JsonMaskingConfig maskingConfig;
     private final TrieNode root;
```
Breus, line 48: Now the KeyMatcher has a `TrieNode` and a `StatefulRadixTrieNode` which is quite confusing. Looking at it for a bit I think one is for JsonPath support and the other isn't? In any case this is something that is not clearly documented (or I couldn't find it) and quite counterintuitive without docs. Maybe we can also rename `TrieNode` to `RadixTrieNode` because the difference between the two is only that one has (JsonPath??) state and the other doesn't? Both seem to be Radix Trie nodes

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -520,4 +578,73 @@ void addUpper(Byte b, PreInitTrieNode child) {
             childrenUpper.put(b, child);
         }
     }
+
+    /**
```
Breus, line 582: This JavaDoc is quite hard to understand without first understanding the full code. Maybe we can rewrite it so it clarifies a couple things? E.g.: what is "the sequence" and "repeated matching"?

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -520,4 +578,73 @@ void addUpper(Byte b, PreInitTrieNode child) {
...
+            this.sequence = this.checkPointSequence = 0;
+        }
+
+        StatefulRadixTrieNode(StatefulRadixTrieNode node) {
+            this.node = this.checkPointNode = node.node;
+            this.sequence = this.checkPointSequence = node.sequence;
+        }
+
+        @Nullable
+        StatefulRadixTrieNode child(byte b) {
```
Breus, line 605: `getChild` maybe? I think this one can also 1 sentence JavaDoc what `child()` means (it returns the trie node with the child currently set as "active node"? )

src/test/java/dev/blaauwendraad/masker/json/InstanceCreationMemoryUsageTest.java
```diff
@@ -1623,7 +1623,7 @@ void realWorldObfuscatedKeysInstanceCreation() {
 
         long memoryBeforeInstanceCreationKb = getCurrentRetainedMemory();
 
-        long memoryLimitKb = 2_000;
```
Breus, line 1626: Very sexy change :-D

src/test/java/dev/blaauwendraad/masker/json/KeyMatcherTest.java
```diff
@@ -255,4 +255,30 @@ private int indexOf(byte[] bytes, char c) {
         }
         return found;
     }
+
+    @Test
+    void printsNicely() {
```
Breus, line 260: Nice, indeed

build.gradle.kts
```diff
@@ -174,12 +174,17 @@ tasks {
...
                 "RedundantThrows",
                 "RemoveUnusedImports",
+                "DefaultCharset",
                 "UnnecessarilyFullyQualified",
                 "UnnecessarilyUsedValue",
                 "UnnecessaryBoxedAssignment",
                 "UnnecessaryBoxedVariable",
                 "UnnecessaryFinal",
                 "UnusedException",
+                "UnusedLabel",
```
Breus, line 184: Where did you find these ones all of a sudden? :-D

Breus (changes requested):

build.gradle.kts
gavlyukovskiy: Mostly looking at them all :D 
https://errorprone.info/bugpatterns

Breus:
@gavlyukovskiy ignore the comments and last commit for now, I am applying the comments myself and want to play around with it a bit more adding lower level unit tests and some JavaDoc

build.gradle.kts
```diff
@@ -174,14 +175,22 @@ tasks {
...
+                "UnusedMethod",
+                "UnusedNestedClass",
+                "UnusedVariable",
                 "WildcardImport",
             )
+            if (DefaultNativePlatform.getCurrentOperatingSystem().isWindows) {
+                disable("MisleadingEscapedSpace") // good stuff
```
Breus, line 192: What's so good about this stuff? 😄

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -105,15 +105,25 @@ public int bufferSize() {
     }
 
     /**
-     * Returns the config for the given key. If no specific config is available for the given key, the default config.
+     * Returns the config for the given key. If no specific config is available for the given key, returns the default config.
      *
      * @param key key to be masked
```
Breus, line 110:
```suggestion
     * @param key the key to be masked
```

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -105,15 +105,25 @@ public int bufferSize() {
...
+     * @return the config for the given key or the default config
      */
     public KeyMaskingConfig getConfig(String key) {
         return targetKeyConfigs.getOrDefault(key, defaultConfig);
     }
 
+    /**
+     * Returns the config for the given key. If no specific config is available for the given key, returns {@code null}.
+     *
+     * @param key key to be masked
```
Breus, line 120:
```suggestion
     * @param key the key to be masked
```

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -32,20 +32,23 @@
...
+ * <p>For masking, we only care whether the key matched or not, so we can use a radix trie to
+ * optimize the look-ups.
  *
- * <p>Further, at initialization time, a case-insensitive trie is created such that any casing
+ * <p>Further, at initialization time, a case-insensitive radix trie is created such that any casing
  * transformations on the looked-up keys during search are avoided.
  *
- * <p>We can also make a trie that looks at bytes instead of characters, so that we can use the
- * bytes and offsets directly in the incoming JSON for comparison and make sure there are no
+ * <p>We can also make a radix trie that looks at bytes instead of characters, so that we can use
```
Breus, line 41:
```suggestion
 * <p>We create a radix trie that looks at bytes instead of characters, so that we can use
```

src/main/java/dev/blaauwendraad/masker/json/JsonPathState.java
```diff
@@ -0,0 +1,103 @@
+package dev.blaauwendraad.masker.json;
+
+import org.jspecify.annotations.Nullable;
+
+import java.util.ArrayDeque;
+
+class JsonPathState {
```
Breus, line 7:
```suggestion
/**
 * Represents the current JSONPath state within the JSON that is being masked. 
 */
class JsonPathState {
```

build.gradle.kts
gavlyukovskiy: It only fails on Windows, so I had to disable it in order to compile code 😅

gavlyukovskiy: It's supposed to check that you only use \s at the end of the line in multiline strings (to preserve whitespaces). Works well on Linux and MacOS.

src/main/java/dev/blaauwendraad/masker/json/JsonPathState.java
```diff
@@ -0,0 +1,103 @@
...
+    private static final KeyMatcher.StatefulRadixTrieNode NULL_NODE = new KeyMatcher.StatefulRadixTrieNode(new KeyMatcher.RadixTrieNode(new byte[0], new byte[0]), 0);
+
+    private final KeyMatcher keyMatcher;
+    private final ArrayDeque<KeyMatcher.StatefulRadixTrieNode> jsonPathSegments = new ArrayDeque<>();
+
+    JsonPathState(KeyMatcher keyMatcher) {
```
donavdey, line 13: I'd call this class "JsonPathTracker" as it also has a lookup logic besides storing the jsonpath state.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -313,211 +351,311 @@ && isEncodedCharacter(bytes, i, endIndex)) {
...
+
+    /**
+     * A helper class for matching against the (radix) trie node.
+     * <p>During matching the node keeps transient matching state - reference to the node {@link #tempNode} and
+     * the {@link #tempPrefixIndex} for keeping track of how much of the common prefix has been matched already.
+     * During matching, this helper class changes the
+     *
+     * <p> After the matching, the node can be reset to the initial state using {@link #reset()}.
+     */
+    static class StatefulRadixTrieNode {
```
donavdey, line 588: How about calling it "RadixTrieNodePrefixPointer"? This name would remove the impression that this is some different kind of a radix trie node, while making it clear that it is a pointer to an index of the prefix represented by the nested trie node.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -313,211 +351,311 @@ && isEncodedCharacter(bytes, i, endIndex)) {
...
+         *
+         * <p>This mutates the current state
+         * prefix matches the byte value, return the existing radix trie node with increased prefix index. Otherwise,
+         * looks if a child node matches the byte value. If neither match, returns {@code null}.
+         *
+         * @param byteValue the "next" byte value to find the matching radix trie node for
+         * @return the current radix trie node or a child of it, if either of them match. Otherwise, returns
+         * {@code null}
+         */
+        boolean performChildLookup(byte byteValue) {
```
donavdey, line 615: The name is a bit confusing as I would expect the method to return a looked up child. How about calling it "descent"? This would indicate that the method changes the state of the object.

Breus (approved):

PR #163, opened 2024-08-04
Implement input/output streaming support

donavdey:
The PR adds support for Input/Output streams.  

~~The implementation has a new type of MaksingState called "BufferedMaskingState". BufferedMaskingState is a wrapper around MaskingState, which got renamed to "ByteArrayMaskingState" as MaskingState now is an interface.~~ 
The impementation enhances MaskingState to support buffering data from input stream and flushing replacement operations into the output stream. Both streamind mode and byte array mode share the same execution path.

BufferedMaskingState reads chunks of size 8192 bytes from the provided input stream and buffers the data in ByteArrayMaskingState. In case the end of the current buffer is reached while the masker processes a JSON value, the current buffer gets extended and everything before the start of the current JSON value gets cut. In case the masker does not process a JSON value, the buffer gets replaced with the next one.  

For writes, BufferedMaskingState flushes prepared replacement operations byte array into provided output stream. In case the data is flushed while the masker is still processing the current JSON value, everything after the start of the current value gets cut in the output byte array.  

The PR adds an assertion to every test that the output of bytes API and streams API is equivalent. There is also a disabled test that creates a ~1GB file and asserts that the masker is able to process it with streams API without going OOM.

The PR also creates StreamsBenchmark JMH suite. Those should only be used to compare performance for different types of streams as they have a specific JMH setup which has some timing measurement implications.

### DIFF
--- a/src/jmh/java/dev/blaauwendraad/masker/json/JsonMaskerBenchmark.java
+++ b/src/jmh/java/dev/blaauwendraad/masker/json/JsonMaskerBenchmark.java
@@ -14,6 +14,8 @@ import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.Warmup;
 import dev.blaauwendraad.masker.json.util.JsonPathTestUtils;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -65,5 +67,10 @@ public class JsonMaskerBenchmark {
     @Benchmark
     public byte[] jsonMaskerBytes(State state) {
         return state.jsonMasker.mask(state.jsonBytes);
+    }
+
+    @Benchmark
+    public void jsonMaskerByteArrayStreams(State state) {
+        state.jsonMasker.mask(new ByteArrayInputStream(state.jsonBytes), new ByteArrayOutputStream());
     }
 }
--- a/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
@@ -6,6 +6,9 @@ import dev.blaauwendraad.masker.json.util.AsciiCharacter;
 import dev.blaauwendraad.masker.json.util.AsciiJsonUtil;
 import org.jspecify.annotations.Nullable;
 
+import java.io.InputStream;
+import java.io.OutputStream;
+
 /**
  * Default implementation of the {@link JsonMasker}.
  */
@@ -17,7 +20,7 @@ final class KeyContainsMasker implements JsonMasker {
     /**
      * The masking configuration for the JSON masking process.
      */
-    private final JsonMaskingConfig maskingConfig;
+    final JsonMaskingConfig maskingConfig;
 
     /**
      * Creates an instance of an {@link KeyContainsMasker}
@@ -39,9 +42,26 @@ final class KeyContainsMasker implements JsonMasker {
      */
     @Override
     public byte[] mask(byte[] input) {
-        try {
-            MaskingState maskingState = new MaskingState(input, !maskingConfig.getTargetJsonPaths().isEmpty());
+        MaskingState maskingState = new MaskingState(input, !maskingConfig.getTargetJsonPaths().isEmpty());
+        return mask(maskingState);
+    }
 
+    /**
+     * Runs masker in a streaming mode.
+     * The masker buffers data from provided input stream in chunks of size 8192 bytes and processes each chunk
+     * sequentially. The output is written into provided output stream after processing each chunk.
+     *
+     * @param inputStream the JSON input stream
+     * @param outputStream masked JSON output stream
+     */
+    @Override
+    public void mask(InputStream inputStream, OutputStream outputStream) {
+        MaskingState maskingState = new MaskingState(inputStream, outputStream, !maskingConfig.getTargetJsonPaths().isEmpty(), maskingConfig.bufferSize());
+        mask(maskingState);
+    }
+
+    private byte[] mask(MaskingState maskingState) {
+        try {
             KeyMaskingConfig keyMaskingConfig = maskingConfig.isInAllowMode() ? maskingConfig.getDefaultConfig() : null;
             if (maskingState.jsonPathEnabled()) {
                 maskingState.expandCurrentJsonPath(keyMatcher.getJsonPathRootNode());
@@ -54,6 +74,7 @@ final class KeyContainsMasker implements JsonMasker {
                     maskingState.next();
                 }
             }
+            maskingState.flushCurrentBuffer();
 
             return maskingState.flushReplacementOperations();
         } catch (ArrayIndexOutOfBoundsException | StackOverflowError e) {
@@ -163,15 +184,17 @@ final class KeyContainsMasker implements JsonMasker {
                 break;
             }
             // In case target keys should be considered as allow list, we need to NOT mask certain keys
-            int openingQuoteIndex = maskingState.currentIndex();
+            maskingState.registerTokenStartIndex();
 
             stepOverStringValue(maskingState);
 
+            int keyStartIndex = maskingState.getCurrentTokenStartIndex();
             int afterClosingQuoteIndex = maskingState.currentIndex();
-            int keyLength = afterClosingQuoteIndex - openingQuoteIndex - 2; // minus the opening and closing quotes
-            maskingState.expandCurrentJsonPath(keyMatcher.traverseJsonPathSegment(maskingState.getMessage(), maskingState.getCurrentJsonPathNode(), openingQuoteIndex + 1, keyLength));
-            KeyMaskingConfig keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), openingQuoteIndex + 1, // plus one for the opening quote
+            int keyLength = afterClosingQuoteIndex - keyStartIndex - 2; // minus the opening and closing quotes
+            maskingState.expandCurrentJsonPath(keyMatcher.traverseJsonPathSegment(maskingState.getMessage(), maskingState.getCurrentJsonPathNode(), keyStartIndex + 1, keyLength));
+            KeyMaskingConfig keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), keyStartIndex + 1, // plus one for the opening quote
                     keyLength, maskingState.getCurrentJsonPathNode());
+            maskingState.clearTokenStartIndex();
             stepOverWhitespaceCharacters(maskingState);
             // step over the colon ':'
             maskingState.next();
@@ -217,12 +240,12 @@ final class KeyContainsMasker implements JsonMasker {
      * @param keyMaskingConfig the {@link KeyMaskingConfig} for the corresponding JSON key
      */
     private void maskString(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
-        maskingState.registerValueStartIndex();
+        maskingState.registerTokenStartIndex();
         stepOverStringValue(maskingState);
 
         keyMaskingConfig.getStringValueMasker().maskValue(maskingState);
 
-        maskingState.clearValueStartIndex();
+        maskingState.clearTokenStartIndex();
     }
 
     /**
@@ -238,12 +261,12 @@ final class KeyContainsMasker implements JsonMasker {
      */
     private void maskNumber(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
         // This block deals with numeric values
-        maskingState.registerValueStartIndex();
+        maskingState.registerTokenStartIndex();
         stepOverNumericValue(maskingState);
 
         keyMaskingConfig.getNumberValueMasker().maskValue(maskingState);
 
-        maskingState.clearValueStartIndex();
+        maskingState.clearTokenStartIndex();
     }
 
     /**
@@ -255,12 +278,12 @@ final class KeyContainsMasker implements JsonMasker {
      * @param keyMaskingConfig the {@link KeyMaskingConfig} for the corresponding JSON key
      */
     private void maskBoolean(MaskingState maskingState, KeyMaskingConfig keyMaskingConfig) {
-        maskingState.registerValueStartIndex();
+        maskingState.registerTokenStartIndex();
         maskingState.incrementIndex(AsciiCharacter.isLowercaseT(maskingState.byteAtCurrentIndex()) ? 4 : 5);
 
         keyMaskingConfig.getBooleanValueMasker().maskValue(maskingState);
 
-        maskingState.clearValueStartIndex();
+        maskingState.clearTokenStartIndex();
     }
 
     /**
--- a/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
@@ -19,6 +19,7 @@ final class KeyContainsMasker implements JsonMasker {
     private final KeyMatcher keyMatcher;
     /**
      * The masking configuration for the JSON masking process.
+     * Package private for unit tests.
      */
     final JsonMaskingConfig maskingConfig;
 
--- a/src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
@@ -15,7 +15,18 @@ import java.util.List;
  * operation.
  */
 final class MaskingState implements ValueMaskerContext {
-    private static final int INITIAL_JSONPATH_STACK_CAPACITY = 16; // an initial size of the jsonpath array
+    private static final int INITIAL_JSONPATH_STACK_CAPACITY = 16; // an initial size of the JsonPath array
+    /**
+     * Defines the maximum size for the buffer used by the streaming API:
+     * {@link JsonMasker#mask(InputStream, OutputStream)}.
+     * <p>
+     * This is a security measure to prevent too much memory being allocated for maliciously crafted JSONs with huge
+     * tokens (keys or values) to consume too much memory.
+     * <p>
+     * The maximum allowed buffer size corresponds to 16MB, which corresponds to a maximum token length of 4 million
+     * characters.
+     */
+    private static final int MAX_BUFFER_SIZE = 16777216;
     private static final String STREAM_READ_ERROR_MESSAGE = "Failed to read from input stream";
     private static final String STREAM_WRITE_ERROR_MESSAGE = "Failed to write to output stream";
 
@@ -24,7 +35,19 @@ final class MaskingState implements ValueMaskerContext {
     private int bufferSize; // size of byte array buffers to be read from the input stream
     private int currentIndex = 0;
     private final List<ReplacementOperation> replacementOperations = new ArrayList<>();
+    /**
+     * The index marking the end of the last replacement operation. Used to determine the start point for the next
+     * replacement operation to ensure they don't overlap.
+     */
     private int lastReplacementEndIndex = 0;
+    /**
+     * The total difference in the replacement operations related to the byte masking process.
+     * This counter aggregates the difference in lengths between the original byte sequences and
+     * their corresponding replacement masks over multiple replacement operations.
+     * <p>
+     * It is used to keep track of the cumulative change in byte count which informs buffer adjustments
+     * during the masking process in order to accommodate the length variations due to replacements.
+     */
     private int replacementOperationsTotalDifference = 0;
     @Nullable private final InputStream inputStream;
     @Nullable private final OutputStream outputStream;
@@ -50,15 +73,14 @@ final class MaskingState implements ValueMaskerContext {
     public MaskingState(InputStream inputStream, OutputStream outputStream, boolean trackJsonPath, int bufferSize) {
         /*
          There is a special optimization for "true", "false" and "null" values. We identify such values by their first
-         character ("t", "f" and "n" respectively) and assume the identified value length. When masker is in allow mode,
-         we may step over these values. In case the buffer size is less than the maximum possible length of such
+         character ("t", "f" and "n" respectively) and assume the identified value length. When the masker is in allow
+         mode, we may step over these values. In case the buffer size is less than the maximum possible length of such a
          "special" value, we end up stepping over the entire buffer. To mitigate that, we force the minimum buffer size
-         to be the maximum possible length, which is 5 (in "false").
+         to be the maximum possible length of such "special" values, which is 5 (in "false").
         */
         if (bufferSize < 5) {
             throw new IllegalArgumentException("Buffer size must be at least 5 bytes");
         }
-
         this.inputStream = inputStream;
         this.outputStream = outputStream;
         this.bufferSize = bufferSize;
@@ -70,6 +92,12 @@ final class MaskingState implements ValueMaskerContext {
         readNextBuffer();
     }
 
+    /**
+     * Advances to the next byte in the message or (Streaming API) buffer, expanding the buffer if necessary.
+     *
+     * @return {@code true} if the current index is within the bounds of the message or if the (Streaming API) buffer
+     * was successfully reloaded and more data is available in the stream, {@code false} otherwise
+     */
     public boolean next() {
         return ++currentIndex < messageLength || reloadBuffer();
     }
@@ -99,6 +127,11 @@ final class MaskingState implements ValueMaskerContext {
      * replacement is done in-place, otherwise a replacement operation is recorded to be performed as a batch using
      * {@link #flushReplacementOperations}.
      *
+     * @param startIndex the start index of the target value in the byte array
+     * @param length     the length of the target value to be replaced
+     * @param mask       the byte array representing the mask
+     * @param maskRepeat the number of times the mask should be repeated
+     * @throws UncheckedIOException if an I/O error occurs while writing to the output stream
      * @see ReplacementOperation
      */
     public void replaceTargetValueWith(int startIndex, int length, byte[] mask, int maskRepeat) {
@@ -188,7 +221,7 @@ final class MaskingState implements ValueMaskerContext {
 
     /**
      * Checks if jsonpath masking is enabled.
-     * @return true if jsonpath masking is enabled, false otherwise
+     * @return {@code true} if JSONPath masking is enabled, {@code false} otherwise
      */
     boolean jsonPathEnabled() {
         return currentJsonPath != null;
@@ -210,7 +243,7 @@ final class MaskingState implements ValueMaskerContext {
     }
 
     /**
-     * Backtracks current jsonpath to the previous segment.
+     * Backtracks current JSONPath to the previous segment.
      */
     void backtrackCurrentJsonPath() {
         if (currentJsonPath != null) {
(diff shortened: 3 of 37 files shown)


gavlyukovskiy:
Without looking much into the changes in the PR
> The implementation has a new type of MaksingState called "BufferedMaskingState". BufferedMaskingState is a wrapper around MaskingState, which got renamed to "ByteArrayMaskingState" as MaskingState now is an interface.

why even do this distinction? The byte array can be always wrapped into a `ByteArrayInputStream` as input and `ByteArrayOutputStream` as output and processed in the "streaming mode". I wouldn't keep two separate implementations of `MaskingState` unless there's a huge performance hit with the byte array wrappers.

donavdey:
> Without looking much into the changes in the PR
> 
> > The implementation has a new type of MaksingState called "BufferedMaskingState". BufferedMaskingState is a wrapper around MaskingState, which got renamed to "ByteArrayMaskingState" as MaskingState now is an interface.
> 
> why even do this distinction? The byte array can be always wrapped into a `ByteArrayInputStream` as input and `ByteArrayOutputStream` as output and processed in the "streaming mode". I wouldn't keep two separate implementations of `MaskingState` unless there's a huge performance hit with the byte array wrappers.

Streaming mode turned out to be more complex than just wrapping the byte arrays into ByteArrayInputStream/ByteArrayOutputStream.
BufferedMaskingState is a wrapper which is primarily responsible for bufferization. Essentially it reads the next byte array from the stream and adjust the pointers in the masking state (i.e. currentIndex and currentValueStartIndex) while taking care of nuances (such as if the masker is currently looping through a JSON value).  
Also I had an idea that someone will pass exceptionally long PipedInputStream or FileInputStream that could not be kept in memory. The current implementation handles them fine, there is a disabled test "LargeStreamTest" that creates 1GB json file and asserts that the masker is able to process it.

donavdey:
> > Without looking much into the changes in the PR
> > > The implementation has a new type of MaksingState called "BufferedMaskingState". BufferedMaskingState is a wrapper around MaskingState, which got renamed to "ByteArrayMaskingState" as MaskingState now is an interface.
> > 
> > 
> > why even do this distinction? The byte array can be always wrapped into a `ByteArrayInputStream` as input and `ByteArrayOutputStream` as output and processed in the "streaming mode". I wouldn't keep two separate implementations of `MaskingState` unless there's a huge performance hit with the byte array wrappers.
> 
> Streaming mode turned out to be more complex than just wrapping the byte arrays into ByteArrayInputStream/ByteArrayOutputStream. BufferedMaskingState is a wrapper which is primarily responsible for bufferization. Essentially it reads the next byte array from the stream and adjust the pointers in the masking state (i.e. currentIndex and currentValueStartIndex) while taking care of nuances (such as if the masker is currently looping through a JSON value). Also I had an idea that someone will pass exceptionally long PipedInputStream or FileInputStream that could not be kept in memory. The current implementation handles them fine, there is a disabled test "LargeStreamTest" that creates 1GB json file and asserts that the masker is able to process it.

Now it occured to me that I could have used BufferedInputStream to make the implementation easier, let me try it.

donavdey:
@gavlyukovskiyt to elaborate on the motivation to create a separate BufferedMaskingState implementation. Initially I tried to implement the streaming mode inside MaskingState, but there were 2 nuances.
The first and the main nuance is the random access of the "message" byte array. There is logic that arbitrarily moves the pointer along the byte array including logic in KeyMatcher trie. The only way to achieve random access in input streams is to use mark/reset/skip methods, but rewriting every byte array random access in these methods turned out to be more cumbersome than I expected (maybe I did not try hard enough though).
The second nuance is code clarity. Every input stream method declares that it throws IOException, so handling it in the same class makes the code in MaskingState (and in KeyMatcher) almost incomprehensible.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -154,15 +178,16 @@ private void visitObject(MaskingState maskingState, @Nullable KeyMaskingConfig p
...
-            int openingQuoteIndex = maskingState.currentIndex();
+            maskingState.registerValueStartIndex();
 
             stepOverStringValue(maskingState);
 
             int afterClosingQuoteIndex = maskingState.currentIndex();
-            int keyLength = afterClosingQuoteIndex - openingQuoteIndex - 2; // minus the opening and closing quotes
-            maskingState.expandCurrentJsonPath(keyMatcher.traverseJsonPathSegment(maskingState.getMessage(), maskingState.getCurrentJsonPathNode(), openingQuoteIndex + 1, keyLength));
-            KeyMaskingConfig keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), openingQuoteIndex + 1, // plus one for the opening quote
+            int keyLength = afterClosingQuoteIndex - maskingState.getCurrentValueStartIndex() - 2; // minus the opening and closing quotes
```
Breus, line 186: This change is very odd. Keylength was the after close index minus the open and now it is after close index minus the `maskingState.getCurrentValueStartIndex`. I think this makes the whole code much less readible to understand that this is because this is the index registered in ` maskingState.registerValueStartIndex()`. It is not the start of a current value, it is the start of the key in this case.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -154,15 +178,16 @@ private void visitObject(MaskingState maskingState, @Nullable KeyMaskingConfig p
...
+            maskingState.expandCurrentJsonPath(keyMatcher.traverseJsonPathSegment(maskingState.getMessage(), maskingState.getCurrentJsonPathNode(), maskingState.getCurrentValueStartIndex() + 1, keyLength));
```
Breus, line 187: I think this is also not super clear now. Can you split this in two statements with some doc explaining what is going on?

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -154,15 +178,16 @@ private void visitObject(MaskingState maskingState, @Nullable KeyMaskingConfig p
...
+            KeyMaskingConfig keyMaskingConfig = keyMatcher.getMaskConfigIfMatched(maskingState.getMessage(), maskingState.getCurrentValueStartIndex() + 1, // plus one for the opening quote
                     keyLength, maskingState.getCurrentJsonPathNode());
+            maskingState.clearValueStartIndex();
```
Breus, line 190: You clear a key start index here, so we should rename this method and add a comment in the beginning to say we mark the start of the key.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
```diff
@@ -39,9 +42,31 @@ final class KeyContainsMasker implements JsonMasker {
...
+     *
+     * @param inputStream the JSON input stream
+     * @param outputStream masked JSON output stream
+     */
+    @Override
+    public void mask(InputStream inputStream, OutputStream outputStream) {
+        MaskingState maskingState = new MaskingState(inputStream, outputStream, !maskingConfig.getTargetJsonPaths().isEmpty());
+        mask(maskingState);
+
+        // in case we are given a json that has trailing characters after the outermost object/array is closed
```
Breus, line 62: I don't fully get why it is only if the character is repeated more than 8129 times why not if there is 8000 bytes of message plus 200 bytes of trailing white spaces. Then this also needs to happen, right?

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -14,10 +18,18 @@
  */
 final class MaskingState implements ValueMaskerContext {
     private static final int INITIAL_JSONPATH_STACK_CAPACITY = 16; // an initial size of the jsonpath array
-    private final byte[] message;
+    private static final String STREAM_READ_ERROR_MESSAGE = "Failed to read input stream";
+    private static final String STREAM_WRITE_ERROR_MESSAGE = "Failed to write to output stream";
+
+    private byte[] message;
+    int messageLength;
+    private int bufferSize = 8192; // size of byte array buffers to be read from the input stream
```
Breus, line 26: I was thinking, wouldn't it be easy to have this changeable (in a package private method) such that you can easily test any combination of overlap over multiple buffers? For example, set the buffer to 100 and have tons of hardcoded tests that have overlaps of buffers just before the colon, just after the colon, in the middle of a key, in the middle of an array, object, etc. etc.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -14,10 +18,18 @@
  */
 final class MaskingState implements ValueMaskerContext {
     private static final int INITIAL_JSONPATH_STACK_CAPACITY = 16; // an initial size of the jsonpath array
-    private final byte[] message;
+    private static final String STREAM_READ_ERROR_MESSAGE = "Failed to read input stream";
```
Breus, line 21:
```suggestion
    private static final String STREAM_READ_ERROR_MESSAGE = "Failed to read from input stream";
```

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -29,13 +41,27 @@ final class MaskingState implements ValueMaskerContext {
...
+        this.messageLength = 0;
+        if (trackJsonPath) {
+            currentJsonPath = new KeyMatcher.TrieNode[INITIAL_JSONPATH_STACK_CAPACITY];
+        }
+        readNextBuffer();
     }
 
     public boolean next() {
-        return ++currentIndex < message.length;
+        return ++currentIndex < messageLength || readNextBuffer();
```
Breus, line 64: For readability of the code it might be nicer to make it like `|| (inputStream != && readNextBuffer()) `. Otherwise you need to navigate into readNextBuffer to see that it will always return false in that case (when you don't have an inputStream)

src/test/java/dev/blaauwendraad/masker/json/LargeStreamTest.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    @AfterAll
+    static void cleanUpFiles() throws IOException {
+        // clean up files
+        Files.deleteIfExists(Path.of(INPUT_FILE_NAME));
+        Files.deleteIfExists(Path.of(EXPECTED_RESULT_FILE_NAME));
+        Files.deleteIfExists(Path.of(ACTUAL_RESULT_FILE_NAME));
+    }
```
gavlyukovskiy, line 85: There's nice `@TempFile` extension from JUnit that does that automatically:
```
@TempFile Path inputFile;
@TempFile Path expectedResultFile;
@TempFile Path actualResultFile;
```

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -250,19 +293,86 @@ private void checkCurrentValueBounds(int index) {
...
+    boolean readNextBuffer() {
+        if (inputStream == null || outputStream == null) {
+            return false;
+        }
+
+        // flush the remaining of the current buffer up to the current value start index into output stream
+        try {
+            int end = currentValueStartIndex == -1 ? messageLength : currentValueStartIndex;
+            outputStream.write(message, lastReplacementEndIndex, end - lastReplacementEndIndex);
+            outputStream.flush();
```
gavlyukovskiy, line 313: Why do we need flush here?

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -250,19 +293,86 @@ private void checkCurrentValueBounds(int index) {
...
+                System.arraycopy(message, currentValueStartIndex, message, 0, currentValueLength);
+            } else {
+                // in case the current value is longer than a quarter of the buffer,
+                // extend the buffer length by a quarter
+                bufferSize += bufferSize >> 2;
+                byte[] extension = new byte[bufferSize];
+
+                // move the current value to the beginning of the extension
+                System.arraycopy(message, currentValueStartIndex, extension, 0, currentValueLength);
+                message = extension;
```
gavlyukovskiy, line 345: I'd rather extend by x2, instead of a quarter. While this is highly unlikely, it's generally how most collections amortize the complexity.

Another question: what happens if extending the buffer once isn't enough?

src/test/java/dev/blaauwendraad/masker/json/MaskingStateTest.java
```diff
@@ -91,4 +92,82 @@ void shouldUseCorrectOffsetWhenThrowingValueMaskerError() {
...
+
+    @Test
+    void shouldReachEndOfStream() {
+        String json = "{\"maskMe:\":\"val\"}";
+
+        MaskingState maskingState = new MaskingState(new ByteArrayInputStream(json.getBytes(StandardCharsets.UTF_8)),
+                new ByteArrayOutputStream(), false);
+
+        Assertions.assertThat(maskingState.messageLength).isEqualTo(17);
+    }
```
gavlyukovskiy, line 172: Maybe I'm missing it, but are there tests to cover these scenarios?
1. Extending buffer when the (matching) key is in between the first quarter and the rest of the buffer
2. Extending buffer when the (matching) key is fully in the last 3/4 of the buffer
3. Extending buffer when the (matching) key is larger than 2 buffers sizes
4. Extending buffer when the (masked) value is in between the first quarter and the rest of the buffer
5. Extending buffer when the (masked) value is fully in the last 3/4 of the buffer
6. Extending buffer when the (masked) value is larger than 2 buffer sizes
7. Using `ValueMaskers.withTextValue` on value that extends the buffer (to verify that all `ValueMaskerContext` methods work correctly)

src/jmh/java/dev/blaauwendraad/masker/json/JsonMaskerBenchmark.java
```diff
@@ -66,4 +68,9 @@ public String jsonMaskerString(State state) {
     public byte[] jsonMaskerBytes(State state) {
         return state.jsonMasker.mask(state.jsonBytes);
     }
+
+    @Benchmark
+    public void jsonMaskerByteArrayStreams(State state) {
+        state.jsonMasker.mask(new ByteArrayInputStream(state.jsonBytes), new ByteArrayOutputStream());
+    }
```
gavlyukovskiy, line 75: Isn't this also covered by the benchmark below?

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -66,9 +92,26 @@ public byte[] getMessage() {
...
+            // write the replacement into the output stream
+            try {
+                // write everything up to the beginning of the current replacement
+                outputStream.write(message, lastReplacementEndIndex, startIndex - lastReplacementEndIndex);
+
+                // write the replacement
+                for (int i = 0; i < maskRepeat; i++) {
+                    outputStream.write(mask);
+                }
+                outputStream.flush();
```
gavlyukovskiy, line 109: Do we actually need `flush` here?

gavlyukovskiy (commented):
Great work! Adde a couple of comments. I think we'll need some more tests to make sure every possible scenario with buffer operations are covered sufficiently.

Also, file benchmarks look a bit sus
```
StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A    ByteArrayStream     ByteArrayStream  thrpt    4    393902.826 ±    13635.863   ops/s
StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A    ByteArrayStream          FileStream  thrpt    4      4326.934 ±      499.007   ops/s
StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A         FileStream     ByteArrayStream  thrpt    4     89939.494 ±      499.991   ops/s
StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A         FileStream          FileStream  thrpt    4      5089.371 ±     2363.523   ops/s
```
Is it because of flushing on every replacement? It would be nice to also update `BaselineBenchmark` with simply writing a json into a file without any masking to see how it compares.

src/test/java/dev/blaauwendraad/masker/json/MaskingStateTest.java
Breus: This is what we had in mind as well, and we think it would be easier if you can configure the buffer size to cover all these cases.

src/jmh/java/dev/blaauwendraad/masker/json/JsonMaskerBenchmark.java
donavdey: It is. I included streaming mode benchmarks here to show how it compares to the rest of the modes with varying masker parameters, while the idea of the benchmark below (`StreamTypeBenchmark`) is to see how streaming mode performs for different types of streams with fixed masker parameters.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Not really. Now as I think of it, a single flush while reloading the buffer makes more sense.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Good idea

> Another question: what happens if extending the buffer once isn't enough?

The current value length is always less than the size of the current buffer. So we keep increasing current buffer size every iteration untill we reach the end of the value.

src/test/java/dev/blaauwendraad/masker/json/LargeStreamTest.java
donavdey: Nice one, didn't know about this

src/test/java/dev/blaauwendraad/masker/json/MaskingStateTest.java
donavdey: Agree, I will make the buffer size package-private and configure it small enough in tests so that I can cover these scenarious.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: It is not bound to the buffer size. Internally, `readNextBuffer` first flushes the old buffer into the output stream and then reads the next one. So the while loop below will basically read the remaining of the input stream and flush them into the output stream in chunks of `bufferSize` length.
So even if there is only 1 whitespace character left in the input stream, the character gets transfered to the output stream.  
Also I agree that `readNextBuffer` is not a good name for a method because it does not imply that the method also flushes the old buffer, I will rename it.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: The idea of the change is not to expose currentIndex anywhere outside of MaskingState so that we are sure that we can safely reset the indexes because there is no one "outside" who remembers old indexes while the next buffer is being read.
It is a good point that this is a start of the key and not the start of the current value. I will rename the method to `registerTokenStartIndex` as this new name reflects both registering value start and key start indexes.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: You are right, it does look cumbersome. I will fix it.

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: Good point, I will do it.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Good point. I think the best way is to make the field package-private and set it to small enough value in unit tests to cover scenarious below.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: I think that also would be slightly better for performance, I will do that.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -250,19 +293,86 @@ private void checkCurrentValueBounds(int index) {
...
+     *
+     * @return true if more data is available in the input stream, false if the end of the stream is reached.
+     */
+    boolean readNextBuffer() {
+        if (inputStream == null || outputStream == null) {
+            return false;
+        }
+
+        // flush the remaining of the current buffer up to the current value start index into output stream
+        try {
```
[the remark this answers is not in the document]
donavdey: Agree, I will break it up into multple smaller methods and give them appropriate names.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: After I remove the flush in "replaceValue" method, this will be the only flush that we do to the output stream. I think it makes sense that we flush "physical" output stream buffer while we read the next "logical" masker buffer.

Breus: Agreed

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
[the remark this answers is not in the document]
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Done

src/test/java/dev/blaauwendraad/masker/json/LargeStreamTest.java
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Done

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Done

src/test/java/dev/blaauwendraad/masker/json/MaskingStateTest.java
donavdey: Done, created a separate test suit for that.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: Done

donavdey:
> Great work! Adde a couple of comments. I think we'll need some more tests to make sure every possible scenario with buffer operations are covered sufficiently.
> 
> Also, file benchmarks look a bit sus
> 
> ```
> StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A    ByteArrayStream     ByteArrayStream  thrpt    4    393902.826 ±    13635.863   ops/s
> StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A    ByteArrayStream          FileStream  thrpt    4      4326.934 ±      499.007   ops/s
> StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A         FileStream     ByteArrayStream  thrpt    4     89939.494 ±      499.991   ops/s
> StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A         FileStream          FileStream  thrpt    4      5089.371 ±     2363.523   ops/s
> ```
> 
> Is it because of flushing on every replacement? It would be nice to also update `BaselineBenchmark` with simply writing a json into a file without any masking to see how it compares.
> 
> Also not sure where the difference between `StreamTypeBenchmark.jsonMaskerStreams` and `JsonMaskerBenchmark.jsonMaskerByteArrayStreams` is coming from:
> 
> ```
> JsonMaskerBenchmark.jsonMaskerByteArrayStreams                           unicode       false         1kb          N/A                     0.1                   N/A                N/A                 N/A  thrpt    4    265071.377 ±     4694.735   ops/s
> ```
> 
> both are 1kb jsons, but throughput is 265k vs 393k, is it because of different masking options?

Removed the flushing on every replacement operation and created a basline benchmark for writing a json into a file. Let's see how it compares now.  
As to ByteArray streams comparisons, StreamType benchmarks always have jsonPath enabled. From the most recent run, the benchmarkss look similar:
`JsonMaskerBenchmark.jsonMaskerByteArrayStreams                           unicode        true         1kb          N/A                     0.1                   N/A                N/A                 N/A  thrpt    4    334683.542 ±     4351.616   ops/s`
`StreamTypeBenchmark.jsonMaskerStreams                                        N/A         N/A         1kb          N/A                     N/A                   N/A    ByteArrayStream     ByteArrayStream  thrpt    4    335748.850 ±    10837.677   ops/s`

src/main/java/dev/blaauwendraad/masker/json/KeyContainsMasker.java
donavdey: I have been thinking about this "while" loop for a while and decided not to transfer anything to the output stream after the outermost JSON object/array closure. I think that makes more sense.

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
donavdey: After the change, `NullAway` complains about dereferencing inputStream inside of "reloadBuffer" method. The thing is that inputStream is a class field and the analyzer does not seem to take into account the fact that this class field is null checked every time before this method is called. Also there are some other instances where "reloadBuffer" is called so that null check outside gets duplicated in 2 more places. So I figured that I keep the null check inside the method.

gavlyukovskiy:
Discussion summary:
- Create ADR to cover streaming decision
  - buffer extension
  - max of 50k characters per token
  - explain what a token is
  - min buffer size == 5 (due to jumping)
  - transferring the remaining buffer after end of the outmost JSON
  - when we need to extend the buffer while we're inside the token - we apply speculative optimization of transferring the token to the beginning of the buffer instead of extending the buffer
- Change all tests to always use the new method that checks both byte array and streaming APIs

src/main/java/dev/blaauwendraad/masker/json/MaskingState.java
```diff
@@ -180,43 +229,51 @@ void backtrackCurrentJsonPath() {
...
     }
 
-    public int getCurrentValueStartIndex() {
-        if (currentValueStartIndex == -1) {
+    boolean isCurrentTokenRegistered() {
+        return currentTokenStartIndex != -1;
+    }
+
+    public int getCurrentTokenStartIndex() {
+        if (!isCurrentTokenRegistered()) {
```
Breus, line 237: Is this possible to trigger with any input to the masker?

Breus (approved):
This PR turned out awesome, I am very happy with the final result! 

Thanks for this major contribution @donavdey!

PR #166, opened 2024-08-24
Improve dependency updates

gavlyukovskiy:
This PR
- updates some mismatched actions
- adds dependabot checks for github actions
- changes to weekly update cheks
- adds (3rd-party) action to check for gradle updates, also uses official gradle validator to verify that wrapper is genuine


Breus (approved):
Very nice improvements :-)

PR #178, opened 2024-09-20
Handle masking of unicode keys for JSONPath

gavlyukovskiy:
Port of #141 for JSONPath. 

Discovered this during implementation of #153, worth fixing it separately to not show up in comparison benchmarks for the trie :)

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
@@ -251,6 +251,16 @@ final class KeyMatcher {
         }
     }
 
+    /**
+     * Searches the trie node by the key offset in the byte array. The node returned might be a prefix,
+     * so {@link TrieNode#endOfWord} needs to be checked additionally for full key matching.
+     *
+     * @param from from which node to do the search, either root node or existing json path node
+     * @param bytes the byte array containing the key to be matched
+     * @param offset offset of the key in the bytes array
+     * @param length length of the key in the bytes array
+     * @return the node if found, {@code null} otherwise.
+     */
     @Nullable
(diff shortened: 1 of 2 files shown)


donavdey (approved):
Looks good to me, nice that now we use the same `searchNode` method in both JSONPath and normal key lookup

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -217,12 +217,12 @@ private void insert(PreInitTrieNode node, String word, boolean negativeMatch) {
             // check JSONPath first, as it's more specific
             // if found - mask with this config
             // if not found - do not mask
-            if (node != null && node.endOfWord && !node.negativeMatch) {
```
gavlyukovskiy, line 220: Removed `!node.negativeMatch` condition because there can be no negative matched in mask mode, only in allow mode.

src/main/java/dev/blaauwendraad/masker/json/KeyMatcher.java
```diff
@@ -252,8 +252,8 @@ private void insert(PreInitTrieNode node, String word, boolean negativeMatch) {
     }
 
     @Nullable
-    private TrieNode searchNode(byte[] bytes, int offset, int length) {
-        TrieNode node = root;
+    private TrieNode searchNode(TrieNode from, byte[] bytes, int offset, int length) {
```
Breus, line 255: Would be nice if this had JavaDoc. Especially why offset and length are relevant in this context is not super clear from just the signature

gavlyukovskiy: Added 👍

PR #184, opened 2024-09-25
[WIP] Improve benchmark pipeline

gavlyukovskiy:
Adding benchmarks pipeline based on this action: https://github.com/marketplace/actions/continuous-benchmark

The current pipeline is to backfill all historical benchmark data, which prompted me to do some interesting stuff with the code, so ignore removing nullaway and adding old jsr305 dependencies, this all is needed to be able to compile latest benchmarks together with the historical code 🙃
I also tried to unify all benchmarks to use the same parameters, so we have something that's consistent.

@Breus @donavdey please review the changes to benchmarks (parameters and json path changes), ignoring all the others for now. If we need any changes to the benchmarks and backfill the historical data, then we will need all these hacks to rerun them.

The benchmarks are pushed into https://gavlyukovskiy.github.io/json-masker-benchmarks/, but I plan to change that to `json-masker` repository, for that we need to enable github-pages for the repository and push into `gh-pages` branch used to render it (and also remove my clear-text PAT 😅). 

The pipeline will need to be configured to only push results from `master` branch and comment on PRs, it has a couple of options like [commenting](https://github.com/benchmark-action/github-action-benchmark?tab=readme-ov-file#alert-comment-on-commit-page) on specific threshold or always, but it requires merging the pipeline into master to test it out. Or we can play with it in a separate repository.


Breus:
@gavlyukovskiy would it be nice if I prepare `json-masker.blaauwendraad.dev` to publish benchmark results and maybe some kind of public documentation page? We can also host our own JavaDoc there :-)

gavlyukovskiy:
@Breus yes, please enable gh pages (some repo setting, I think), you will need to commit empty .nojekyll into the root of gh-pages branch and then I can transfer everything from my repoaitory into that one.

 I also had the same thoughts about javadoc 😁 I think we could do that, but also not much is wrong with our current service. In any case, we can use `/benchmarks` for the benchmarks and leave a room for something else later.

src/jmh/java/dev/blaauwendraad/masker/json/BenchmarkUtils.java
```diff
@@ -15,6 +22,25 @@
 
 public class BenchmarkUtils {
 
+    public static final long STATIC_RANDOM_SEED = 1285756302517652226L;
+    public static final int TARGET_KEYS_SIZE = 2385;
```
donavdey, line 26: Very specific

gavlyukovskiy: It's the number of keys we have in target_keys.json, I just needed the precise number somewhere

src/jmh/java/dev/blaauwendraad/masker/json/BenchmarkUtils.java
```diff
@@ -64,9 +90,72 @@ public static String randomJson(Set<String> targetKeys, String jsonSize, String
...
+            } catch (IllegalArgumentException ignore) {
+                disambiguated.remove(current);
+            }
+        }
+        if (disambiguated.size() == jsonPaths.size()) {
+            return disambiguated;
+        }
+        jsonPaths = disambiguated;
+    }
+}
```
donavdey, line 160:
```suggestion
    public static Set<String> disambiguate(Set<String> jsonPaths) {
        while (true) {
            Set<String> disambiguated = new HashSet<>(jsonPaths);
            List<String> jsonPathKeys = new ArrayList<>(jsonPaths);
            Collections.sort(jsonPathKeys);
            JsonPathParser jsonPathParser = new JsonPathParser();
            for (int i = 1; i < jsonPathKeys.size(); i++) {
                String current = jsonPathKeys.get(i - 1);
                String next = jsonPathKeys.get(i);
                try {
                    jsonPathParser.checkAmbiguity(Set.of(jsonPathParser.parse(current), jsonPathParser.parse(next)));
                } catch (IllegalArgumentException ignore) {
                    disambiguated.remove(current);
                }
            }
            if (disambiguated.size() == jsonPaths.size()) {
                return disambiguated;
            }
            jsonPaths = disambiguated;
        }
    }
```

donavdey: Just formatting. I tried to break that `while(true)` loop, but the problem turned out to be more elaborate. If we have [A, B, C, D] json paths and A, B is not ambiguate and B,C is not ambiguate but C,D is ambiguate, we need to check again if B,D and A,D are ambiguate. 
We could somehow sort json paths by their "ambiguity potential", but that's very nuanced and while(true) loop works fine anyway.

donavdey, line 160:
```suggestion
    public static Set<String> disambiguate(Set<String> jsonPaths) {
        List<String> jsonPathKeys = new ArrayList<>(jsonPaths);
        Collections.sort(jsonPathKeys);
        JsonPathParser jsonPathParser = new JsonPathParser();
        for (int i = 1; i < jsonPathKeys.size(); i++) {
            String current = jsonPathKeys.get(i - 1);
            String next = jsonPathKeys.get(i);
            try {
                jsonPathParser.checkAmbiguity(Set.of(jsonPathParser.parse(current), jsonPathParser.parse(next)));
            } catch (IllegalArgumentException ignore) {
                jsonPathKeys.remove(next);
                i--;
            }
        }
        return new HashSet<>(jsonPathKeys);
    }
```

donavdey:
It also probably makes sense to remove `src/jmh/benchmark-history` folder as the history will be stored somewhere else.

gavlyukovskiy:
@donavdey yea, it's a long time irrelevant at this point

Breus:
Since we don't really have new feature requests, in our vision the library is feature-complete, there is no point in improving benchmark set-up. We might do this later and then we can reopen the issue.

PR #185, opened 2024-09-29
Override next() and endOfJson() methods because JIT C2 keeps de-optimizing them

donavdey:
@gavlyukovskiy's [benchmarks chart](https://gavlyukovskiy.github.io/json-masker-benchmarks/) shows that streaming mode PR degrades performance of byte array mode by ~30%. Upon further examination, it turned out that JIT C2 de-optimizes `next()` and `endOfJson()` methods (the most performance critical methods) because of a new `reloadBuffer()` condition. Logically, this condition is never triggered in byte array mode, but apparently the condition invalidates some of JIT assumptions and previous optimizations.  
This PR decouples buffering logic from MaskingState to keep JIT optimizations for `next()` and `endOfJson()` intact for byte array API.

The preformance chart:  
image

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/BufferedMaskingState.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/BufferedMaskingState.java
@@ -1,0 +1,182 @@
+package dev.blaauwendraad.masker.json;
+
+import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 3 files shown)


gavlyukovskiy (approved):
Nice, so after all you went with decoupling them 😀 To me, that now looks much better

PR #216, opened 2025-01-26
Cleanup old json path code paths

gavlyukovskiy:
Saw these lines not covered, turns out because they're never used :)


donavdey (approved):
Looks good to me!

PR #241, opened 2025-05-06
Update dependabot.yml to add reviewer automatically and change interval to monthly

Breus:


Breus:
@gavlyukovskiy since we only have test dependencies I think doing it monthly and automatically having a reviewer assigned can reduce quite a bit of toil (mails). I will also check for grouping options

Issue #318, opened 2026-04-13
ALLOW mode does not check for masked JSON Paths ambiguity.


andrebrait:
When in ALLOW mode, the JSONPaths added via `maskJsonPaths(String, KeyMaskingConfig)` are not checked for ambiguity because the check only checks the `targetJsonPaths` map:

```java
        public JsonMaskingConfig build() {
            JSON_PATH_PARSER.checkAmbiguity(targetJsonPaths);  // <-- no checking for the key configs
            return new JsonMaskingConfig(this);
        }
```

I have never run into any issues because of this, but it looks like an oversight.

PR #319, opened 2026-04-17
Allow custom masking config without defining allowed keys

### LINKED ISSUE #317: JSONPath-specific masking config gets ignored if the list of allowed JSONPaths is empty
By setting the masking config to ALLOW mode, calling `allowJsonPaths(String...)`, calls to `maskJsonPaths(String, KeyMaskingConfig)` cause the resulting specific masking config to be added to `targetKeyConfigs`. 

While both `targetJsonPaths` and `targetKeyConfigs` are inserted into the matcher trie correctly, the `KeyContainsMatcher` never calls `KeyMatcher#getMaskConfigIfMatched(byte[], int, int, RadixTriePointer, RadixTriePointer)` unless the `targetJsonPaths` is not empty. 

This means it's currently impossible to have a config to achieve "mask everything with the default masker, but use these masking config for these keys".

I'm currently working around it by adding a single key to the allowed JSON paths that is impossible to exist (e.g. `("__a" + UUID.randomUUID()).replace("-", "")`) and then my custom masking configs for the other keys work as expected.

So this works:

```java
builder.allowJsonPaths(myImpossibleToExistKey).maskJsonPaths("$.specific.key", myCustomConfig).build();
```

but this does not:

```java
builder.allowJsonPaths().maskJsonPaths("$.specific.key", myCustomConfig).build();
```

andrebrait:
Fixes #317

### DIFF
--- a/src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
+++ b/src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
@@ -9,6 +9,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.jspecify.annotations.Nullable;
 
 /** Contains the JSON masker configurations. */
@@ -32,17 +33,19 @@ public final class JsonMaskingConfig {
 
     private final KeyMaskingConfig defaultConfig;
     private final Map<String, KeyMaskingConfig> targetKeyConfigs;
+    private final Map<String, KeyMaskingConfig> targetJsonPathConfigs;
 
     JsonMaskingConfig(JsonMaskingConfig.Builder builder) {
         if (builder.targetKeyMode == null) {
             throw new IllegalArgumentException("No keys were requested to mask or allow");
         }
         this.targetKeyMode = builder.targetKeyMode;
-        this.targetKeys = builder.targetKeys;
+        this.targetKeys = builder.allowedKeys;
         this.targetJsonPaths = builder.targetJsonPaths;
         this.caseSensitiveTargetKeys = builder.caseSensitiveTargetKeys != null && builder.caseSensitiveTargetKeys;
         this.defaultConfig = builder.defaultConfigBuilder.build();
         this.targetKeyConfigs = builder.targetKeyConfigs;
+        this.targetJsonPathConfigs = builder.targetJsonPathConfigs.entrySet().stream().collect(Collectors.toMap(e -> e.getKey().toString(), Map.Entry::getValue));
     }
 
     /**
@@ -116,6 +119,16 @@ public final class JsonMaskingConfig {
         return targetKeyConfigs.get(key);
     }
 
+    /**
+     * Returns the config for the given JSON path. If no specific config is available for the given path, returns {@code null}.
+     *
+     * @param path the JSON path to be masked
+     * @return the config for the given key
+     */
+    public @Nullable KeyMaskingConfig getJsonPathConfig(String path) {
+        return targetJsonPathConfigs.get(path);
+    }
+
     public KeyMaskingConfig getDefaultConfig() {
         return defaultConfig;
     }
@@ -129,18 +142,27 @@ public final class JsonMaskingConfig {
         return Collections.unmodifiableMap(targetKeyConfigs);
     }
 
+    /**
+     * Returns a map with all masking configs per JSON path.
+     *
+     * @return masking configs per JSON path
+     */
+    public Map<String, KeyMaskingConfig> getJsonPathConfigs() {
+        return Collections.unmodifiableMap(targetJsonPathConfigs);
+    }
+
     @Override
     public String toString() {
         return String.format(
-                "targetKeys=%s,%ntargetJsonPaths=%s,%ntargetKeyMode=%s,%ncaseSensitiveTargetKeys=%s,%ndefaultConfig=%s,%ntargetKeyConfigs=%s%n",
-                targetKeys, targetJsonPaths, targetKeyMode, caseSensitiveTargetKeys, defaultConfig, targetKeyConfigs);
+                "targetKeys=%s,%ntargetJsonPaths=%s,%ntargetKeyMode=%s,%ncaseSensitiveTargetKeys=%s,%ndefaultConfig=%s,%ntargetKeyConfigs=%s,%ntargetJsonPathConfigs=%s%n",
+                targetKeys, targetJsonPaths, targetKeyMode, caseSensitiveTargetKeys, defaultConfig, targetKeyConfigs, targetJsonPathConfigs);
     }
 
     /** Builder to create {@link JsonMaskingConfig} instances using the builder pattern. */
     public static class Builder {
         private static final JsonPathParser JSON_PATH_PARSER = new JsonPathParser();
 
-        private final Set<String> targetKeys = new HashSet<>();
+        private final Set<String> allowedKeys = new HashSet<>();
         private final Set<JsonPath> targetJsonPaths = new HashSet<>();
 
         @Nullable
@@ -151,6 +173,7 @@ public final class JsonMaskingConfig {
 
         private final KeyMaskingConfig.Builder defaultConfigBuilder = KeyMaskingConfig.builder();
         private final Map<String, KeyMaskingConfig> targetKeyConfigs = new HashMap<>();
+        private final Map<JsonPath, KeyMaskingConfig> targetJsonPathConfigs = new HashMap<>();
 
         private Builder() {}
 
@@ -218,13 +241,13 @@ public final class JsonMaskingConfig {
                         + " to customize masking for specific keys in ALLOW mode, use"
                         + " maskKeys that accepts KeyMaskingConfig");
             }
-            if (targetKeys.contains(key) || targetKeyConfigs.containsKey(key)) {
+            if (allowedKeys.contains(key) || targetKeyConfigs.containsKey(key)) {
                 throw new IllegalArgumentException(String.format("Duplicate key '%s'", key));
             }
             // in ALLOW mode this method can be used to set a specific masking config for a key
             if (targetKeyMode != TargetKeyMode.ALLOW) {
                 targetKeyMode = TargetKeyMode.MASK;
-                targetKeys.add(key);
+                allowedKeys.add(key);
             }
             if (config != null) {
                 targetKeyConfigs.put(key, config);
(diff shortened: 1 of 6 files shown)


andrebrait:
I'll take a look and fix it. Weirdly enough, it all passed locally.

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -129,18 +142,27 @@ public Map<String, KeyMaskingConfig> getKeyConfigs() {
...
+                "targetKeys=%s,%ntargetJsonPaths=%s,%ntargetKeyMode=%s,%ncaseSensitiveTargetKeys=%s,%ndefaultConfig=%s,%ntargetKeyConfigs=%s,%ntargetJsonPathConfigs=%s%n",
+                targetKeys, targetJsonPaths, targetKeyMode, caseSensitiveTargetKeys, defaultConfig, targetKeyConfigs, targetJsonPathConfigs);
     }
 
     /** Builder to create {@link JsonMaskingConfig} instances using the builder pattern. */
     public static class Builder {
         private static final JsonPathParser JSON_PATH_PARSER = new JsonPathParser();
 
-        private final Set<String> targetKeys = new HashSet<>();
+        private final Set<String> allowedKeys = new HashSet<>();
```
Breus: @andrebrait I don't think this can be called `allowedKeys` it is only allowed in allow mode, they are target keys and have a `TargetKeyMode`

andrebrait: Probably an accidental rename. I'll fix it.

andrebrait: Fixed in the latest revision

src/main/java/dev/blaauwendraad/masker/json/path/JsonPathParser.java
```diff
@@ -88,21 +87,22 @@ private List<String> parseSegments(String literal) {
         for (int i = 2; i < literal.length() - 1; i++) {
             char symbol = literal.charAt(i);
             char nextSymbol = literal.charAt(i + 1);
-            if (symbol == '.' || (symbol == '[' && segment.length() != 0)) {
+            if (symbol == '.' || (symbol == '[' && !segment.isEmpty())) {
                 segments.add(segment.toString());
-                segment = new StringBuilder();
+                segment.setLength(0);
```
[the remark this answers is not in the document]
andrebrait: Fixed in the latest revision

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -306,7 +329,7 @@ private void addMaskJsonPath(String jsonPath, @Nullable KeyMaskingConfig config)
                 targetJsonPaths.add(parsed);
             }
             if (config != null) {
-                targetKeyConfigs.put(parsed.toString(), config);
+                targetJsonPathConfigs.put(parsed, config);
             }
```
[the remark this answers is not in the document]
andrebrait: Fixed in the latest revision

src/main/java/dev/blaauwendraad/masker/json/path/JsonPathParser.java
```diff
@@ -88,21 +87,22 @@ private List<String> parseSegments(String literal) {
...
         }
-        if (literal.charAt(literal.length() - 1) != ']' && literal.charAt(literal.length() - 1) != '.') {
-            segment.append(literal.charAt(literal.length() - 1));
+        char lastChar = literal.charAt(literal.length() - 1);
+        if (lastChar != ']' && lastChar != '.') {
+            segment.append(lastChar);
         }
-        if (segment.length() != 0 || literal.endsWith("[]")) {
+        if (!segment.isEmpty() || literal.endsWith("[]")) {
             segments.add(segment.toString());
```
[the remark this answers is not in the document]
andrebrait: Fixed in the latest revision

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -116,6 +119,16 @@ public KeyMaskingConfig getConfig(String key) {
         return targetKeyConfigs.get(key);
     }
 
+    /**
+     * Returns the config for the given JSON path. If no specific config is available for the given path, returns {@code null}.
+     *
+     * @param path the JSON path to be masked
+     * @return the config for the given key
```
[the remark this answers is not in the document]
andrebrait: Fixed in the latest revision

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -397,6 +418,19 @@ public Builder caseSensitiveTargetKeys() {
...
+         *
+         * @return the builder instance
+         * @see dev.blaauwendraad.masker.json.config.KeyMaskingConfig
+         */
+        public Builder maskWith(KeyMaskingConfig maskingConfig) {
+            defaultConfigBuilder.maskStringsWith(maskingConfig.getStringValueMasker())
+                    .maskBooleansWith(maskingConfig.getBooleanValueMasker())
+                    .maskNumbersWith(maskingConfig.getNumberValueMasker());
+            return this;
+        }
```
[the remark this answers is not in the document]
andrebrait: Fixed in the latest revision

andrebrait:
@copilot please fix the merge conflicts in this pull request.

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -297,7 +328,7 @@ private void addMaskJsonPath(String jsonPath, @Nullable KeyMaskingConfig config)
                         + " maskJsonPaths that accepts KeyMaskingConfig");
             }
             JsonPath parsed = JSON_PATH_PARSER.parse(jsonPath);
-            if (targetJsonPaths.contains(parsed) || targetKeyConfigs.containsKey(parsed.toString())) {
+            if (targetJsonPaths.contains(parsed) || targetJsonPathConfigs.containsKey(parsed)) {
```
Breus, line 345: This check used to also cover the key configs, and now that JSONPath configs live in their own map neither side sees the other. `addMaskKey` still only looks at `targetKeys` / `targetKeyConfigs`, so this slips through:

```java
JsonMaskingConfig.builder()
        .maskJsonPaths("$.a", configA)
        .maskKeys("$.a", configB)
        .build();
```

On master that throws `Duplicate key '$.a'`. Here it builds fine, and since `KeyMatcher#insert` resolves `getKeyConfig` before falling back to `getJsonPathConfig`, `configA` is silently dropped — the trie only ever gets `configB`. Same thing in the other order, just with the "Duplicate JSONPath" message instead.

Granted, you have to write a JSONPath-looking string into `maskKeys` to hit this, so it won't bite many people. But quietly ignoring one of two configs is the same failure mode as #318, so I'd rather keep it loud. Could you have both `addMaskKey` and `addMaskJsonPath` check both maps?

src/test/java/dev/blaauwendraad/masker/json/CustomKeyMaskingConfigTest.java
```diff
@@ -496,4 +496,37 @@ void maskKeysWithSpecificConfigInAllowMode() {
                 }
                 """);
     }
+
+    @Test
+    void maskKeysWithSpecificConfigInAllowModeWithoutDefaultAllowedKey() {
+        var jsonMasker = JsonMasker.getMasker(JsonMaskingConfig.builder()
+                .allowJsonPaths()
+                .maskKeys(
+                        "maskedLikeEpsteinFiles",
```
Breus, line 505: Epstein masking 🤣 🤣

Breus: @gavlyukovskiy maybe we can sell the masker to MAGA 😏

src/test/java/dev/blaauwendraad/masker/json/CustomKeyMaskingConfigTest.java
```diff
@@ -496,4 +496,37 @@ void maskKeysWithSpecificConfigInAllowMode() {
                 }
                 """);
     }
+
+    @Test
+    void maskKeysWithSpecificConfigInAllowModeWithoutDefaultAllowedKey() {
```
Breus, line 501: Could we get a bit more around this? The change isn't only "path configs are now stored separately" — it flips when `KeyContainsMasker` creates a `JsonPathTracker` at all, so any ALLOW-mode config carrying a path config now goes down the path-tracking branch instead of the flat one. That's a decent behavioural change to land on a single test.

The one thing I'd really like covered is a nested path. `$.maskedPathLikeEpsteinFiles` is a single segment, so it matches on the first key and never exercises tracking as you descend into objects. Something like `$.a.b.secret`, with a sibling `$.a.b.other` left alone, would tell us a lot more.

Beyond that, two cases I think are worth pinning down:

- a path that appears in both `allowJsonPaths(...)` and `maskJsonPaths(path, config)` — I want to know what wins, and that it stays that way
- path configs in MASK mode, to confirm the new tracker condition didn't shift anything on the other side

src/main/java/dev/blaauwendraad/masker/json/config/JsonMaskingConfig.java
```diff
@@ -415,6 +440,21 @@ public Builder caseSensitiveTargetKeys() {
             return this;
         }
 
+        /**
+         * Mask all values according to the provided {@link KeyMaskingConfig}.
+         *
+         * @return the builder instance
+         * @see dev.blaauwendraad.masker.json.config.KeyMaskingConfig
+         */
+        public Builder maskWith(KeyMaskingConfig maskingConfig) {
```
Breus, line 449: Nice little addition, I like having this as a one liner instead of three separate builder calls.

One thing though: it slipped in fairly quietly. It isn't in the PR description, there's no test for it, and it's missing from the README where the other builder methods live. Since this is public API on a library we're stuck with it once it ships, so I'd rather give it a proper introduction. Could you add a test and a README entry, and mention it in the description? #317 doesn't ask for it, so it's easy to scroll past in review.

One design thought while we're here. Right now it copies the three maskers out of `KeyMaskingConfig` by hand. That's complete today, but if someone adds a fourth field to `KeyMaskingConfig` later, this method will quietly stop copying it. No compile error, no failing test, just a config that doesn't fully apply. Would it make sense to move the copying into `KeyMaskingConfig` itself, something like `KeyMaskingConfig.Builder.from(config)`, so the two can't drift apart?

Also totally fine by me if you'd rather split this out into its own PR and keep this one focused on #317. Whatever's easier for you.

